PR #578, opened 2024-03-05
[feat] Update modals to take up full screen in mobile view

kl-cb:
**What changed? Why?**

- Update voter registration flow to use shared form layout components
- Update styling to make modals take up the full screen in mobile view

**UI changes**

_Mobile Web_

| Old 👴         | New 👶         |
| -------------- | -------------- |
| www standwithcrypto org_(Pixel 7) | swc-web-git-kaerah-mobile-modals-full-screen-stand-with-crypto vercel app_(Pixel 7)|
|www standwithcrypto org_(Pixel 7) (1)|localhost_3000_(Pixel 7) (5)|

**Notes to reviewers**

**How has it been tested?**

- [x] Locally
- [x] Development
- [ ] Unit test
- [ ] Functional test

**Change management**
type=routine 
risk=low 
impact=sev5


coinbasemar:
@kl-cb agree with @travisbloom-cb that it feels buggy having the CTAs float in the middle. Another part of this work would likely be appending the CTAs to the bottom of the screen (with 32px spacing from bottom). We'd also want to give more breathing room at the top of the screen. I'd suggest another 32px spacing between the H1 and the close/back buttons. Here's a mock for spacing reference: https://www.figma.com/file/r9ilZQ4QMRpjefdQFq93Az/SWC-SMS?type=design&node-id=1-1752&mode=design&t=J9cbMVzagg0rkfYr-4

travisbloom-cb:
suggested eng notes on expanded scope for when we pick this up:
- we'll prob want to delete DialogHeader, DialogFooter, DialogTitle, DialogDescription in the dialog component file
- We'll want to add a new "DialogFooterCTA" and corresponding "DialogBody" component. The body component will be used in conjunction with the footer CTA component anytime we have modals where, on mobile, we need the CTA to be flex bounded to the bottom. We'll prob want props on these `hasOverflowContent` (or seperate overflow variant components) that will apply shadow styles and overflow scroll to content (see email modal for example)
- Engineer will need to audit all our modals to apply these new components/styling
- @coinbasemar will need to audit all these flows in a preview branch

travisbloom-cb:
Another thing to be careful about - we have true "dialogs" and we have "deeplink" ones where it just looks like a dialog because of our styling (we do this so that we can pregenerate the "opened" dialog state for some of our deeplinked actions). We need to QA this for both of these

See example of a "deeplink" dialog [here](https://github.com/Stand-With-Crypto/swc-web/blob/642-support-contacting-multiple-politicians-across-the-site-instead-of-just-congressperson/src/app/%5Blocale%5D/(homepageDialogDeeplink)/action/voter-registration/page.tsx#L12)

src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx
```diff
@@ -3,14 +3,20 @@ import { UserActionFormCallCongresspersonDeeplinkWrapper } from '@/components/ap
...
 export const revalidate = SECONDS_DURATION.HOUR
 export const dynamic = 'error'
 
 export default function UserActionCallCongresspersonDeepLink({ params }: PageProps) {
   return (
     <HomepageDialogDeeplinkLayout pageParams={params}>
-      <div className={dialogContentPaddingStyles}>
+      <div
+        className={cn(
+          'flex flex-col items-center justify-center max-md:h-full',
```
Comment (travisbloom-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx:16]: i did NOT know about [this](https://tailwindcss.com/docs/responsive-design#targeting-a-breakpoint-range) 🤯 

cc @lucasrmp-cb @benson-budiman-cb

src/components/app/userActionFormEmailCongressperson/index.tsx
```diff
@@ -237,19 +237,6 @@ export function UserActionFormEmailCongressperson({
                 )}
               />
             </div>
-            <FormField
```
travisbloom-cb, line 240: pull the latest from main and make sure it alings with those changes. I think this is a mistake

src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx
Comment (paulovictor-cb) [reply src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx:16]: @kl-cb That found it out! I learned it with her code as well 👏🏻

src/components/app/userActionFormEmailCongressperson/index.tsx
travisbloom-cb: JK, saw this was intential feedback from @coinbasemar

travisbloom-cb (approved):
Theres some [weird behavior on iOS for our deeplinks](https://github.com/orgs/Stand-With-Crypto/projects/1/views/10?pane=issue&itemId=60569612), but this isn't really a bug introduced by your PR, more just something that existed before and is now more pronounced. 

Created a separate ticket so we dont block this work. We should prob do as a fast follow.

PR #612, opened 2024-03-07
[feature] add Sentry alerts for low Base ETH balances for our airdropping wallets

benson-budiman-cb:
**What changed? Why?**

This is a simple PR that adds Sentry alerts for low Base ETH balances for our airdropping wallets. Why? Because we need a robust way to let the team know if we need to (manually) refund our airdropping wallets.

**How has it been tested?**

Triggered cron job locally.

- [X] Locally
- [ ] Development
- [ ] Unit test
- [ ] Functional test

**Change management**
type=routine 
risk=low 
impact=sev5


src/inngest/functions/monitorBaseETHBalances.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { inngest } from '@/inngest/inngest'
+import { getBaseETHBalances } from '@/utils/server/basescan/getBaseETHBalances'
+import { prettyLog } from '@/utils/shared/prettyLog'
+
+export const MONITOR_BASE_ETH_BALANCES_INNGEST_FUNCTION_ID = 'monitor-base-eth-balances'
+const MONITOR_BASE_ETH_BALANCES_INNGEST_RETRY_LIMIT = 5
+
+const CRYPTO_ADDRESSES = [
+  '0xc2A6116e9a1f9aDD1Bb87EEF308f216Bb0304c38',
```
travisbloom-cb, line 12: we can use env vars here right?

src/inngest/functions/monitorBaseETHBalances.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+}
+
+export const monitorBaseETHBalances = inngest.createFunction(
+  {
+    id: MONITOR_BASE_ETH_BALANCES_INNGEST_FUNCTION_ID,
+    retries: MONITOR_BASE_ETH_BALANCES_INNGEST_RETRY_LIMIT,
+    onFailure: onFailureMonitorBaseETHBalances,
+  },
+  { cron: '*/10 * * * *' }, // Every 10 minutes.
```
travisbloom-cb, line 33: can we limit this to only run on prod?

travisbloom-cb (changes requested):

src/inngest/functions/monitorBaseETHBalances.ts
benson-budiman-cb: yes we can - let me make that change

src/inngest/functions/monitorBaseETHBalances.ts
benson-budiman-cb: yup

src/inngest/functions/monitorBaseETHBalances.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { inngest } from '@/inngest/inngest'
+import { getBaseETHBalances } from '@/utils/server/basescan/getBaseETHBalances'
+import { LEGACY_NFT_DEPLOYER_WALLET, SWC_DOT_ETH_WALLET } from '@/utils/server/nft/constants'
+import { prettyLog } from '@/utils/shared/prettyLog'
+
+export const MONITOR_BASE_ETH_BALANCES_INNGEST_FUNCTION_ID = 'monitor-base-eth-balances'
+const MONITOR_BASE_ETH_BALANCES_INNGEST_RETRY_LIMIT = 5
+
+const LOW_ETH_BALANCE_THRESHOLD = Number(process.env.LOW_ETH_BALANCE_THRESHOLD) ?? 0.25
```
kl-cb, line 12: `Number(process.env.LOW_ETH_BALANCE_THRESHOLD) || 0.25`
`??` will only evaluate to 0.25 if `Number(process.env.LOW_ETH_BALANCE_THRESHOLD)` is null or undefined.

src/inngest/functions/monitorBaseETHBalances.ts
```diff
@@ -0,0 +1,56 @@
...
+    const baseETHBalances = await step.run(
+      'monitor-base-eth-balances.get-base-eth-balances',
+      async () => {
+        return await getBaseETHBalances([SWC_DOT_ETH_WALLET, LEGACY_NFT_DEPLOYER_WALLET])
+      },
+    )
+
+    for (const cryptoAddress of baseETHBalances.result) {
+      // Divide by 10^18 to get the balance in ETH.
+      const ethBalance = Number(cryptoAddress.balance) / 10 ** 18
```
kl-cb, line 40: no magic numbers. can we extract `10 ** 18` to variable like `ETH_BASE_UNIT_WEI`

src/utils/server/basescan/getBaseETHBalances.ts
```diff
@@ -0,0 +1,40 @@
...
+        'Content-Type': 'application/json',
+      },
+    })
+    return (await httpResp.json()) as GetBaseETHBalancesResponse
+  } catch (error) {
+    Sentry.captureException(error, {
+      tags: { domain: 'getBaseETHBalances' },
+      extra: { url },
+    })
+    throw error
```
kl-cb, line 38: Do we need to throw an error here? If so, can we wrap the call to this in a try/catch or do we expect the flow to be unrecoverable?

src/inngest/functions/monitorBaseETHBalances.ts
benson-budiman-cb: @kl-cb I want to evaluate to `0.25` if the envionment varaiable is null, undefined, or empty string. Should I instead do something like:

```typescript
const LOW_ETH_BALANCE_THRESHOLD = process.env.LOW_ETH_BALANCE_THRESHOLD
  ? Number(process.env.LOW_ETH_BALANCE_THRESHOLD)
  : 0.25
```

src/utils/server/basescan/getBaseETHBalances.ts
benson-budiman-cb: @kl-cb I am throwing an error here so that the Inngest function knows to re-attempt a specific step.

But now that I am thinking about the cron job a little more... given the pretty frequent cadence of the cron job, I think it's better if I just not try to retry any specific steps and just throw a nonretriable error for the whole Inngest function.

src/inngest/functions/monitorBaseETHBalances.ts
kl-cb: There's a case where `LOW_ETH_BALANCE_THRESHOLD` will be `NaN`. For example, if `process.env.LOW_ETH_BALANCE_THRESHOLD` is 'abcd' then it will try to do `Number('abcd')` which will evaluate to NaN.

You can just do `const LOW_ETH_BALANCE_THRESHOLD = Number(process.env.LOW_ETH_BALANCE_THRESHOLD) || 0.25`

src/utils/server/basescan/getBaseETHBalances.ts
kl-cb: Gotcha, didn't know it would automatically do a retry if it errors out!

src/inngest/functions/monitorBaseETHBalances.ts
benson-budiman-cb: Ooo okay gotcha

src/app/api/inngest/route.ts
```diff
@@ -13,5 +15,6 @@ export const { GET, POST, PUT } = serve({
     emailRepViaCapitolCanaryWithInngest,
     airdropNFTWithInngest,
     cleanupPostalCodesWithInngest,
+    ...(NEXT_PUBLIC_ENVIRONMENT === 'production' ? [monitorBaseETHBalances] : []),
```
travisbloom-cb, line 18: np: we might want fn's like this to exist in testing so we can manually trigger. Might want the conditional to sit on the cron config itself

travisbloom-cb (approved):

src/app/api/inngest/route.ts
benson-budiman-cb: oh 🤦 yeah that would be much better - updating the PR

src/inngest/functions/monitorBaseETHBalances.ts
```diff
@@ -0,0 +1,67 @@
...
+  {
+    id: MONITOR_BASE_ETH_BALANCES_INNGEST_FUNCTION_ID,
+    retries: MONITOR_BASE_ETH_BALANCES_INNGEST_RETRY_LIMIT,
+    onFailure: onFailureMonitorBaseETHBalances,
+  },
+  {
+    ...(NEXT_PUBLIC_ENVIRONMENT === 'production'
+      ? { cron: '*/10 * * * *' }
+      : { event: MONITOR_BASE_ETH_BALANCES_INNGEST_EVENT_NAME }),
+  }, // Every 10 minutes.
```
benson-budiman-cb, line 37: Inngest requires either `cron` or `event`.
- We will use `cron` in production (which can still be triggered manually)
- Otherwise, we use `event`

travisbloom-cb (approved):

PR #617, opened 2024-03-07
Support "verified email sessions" users who aren't logged in for navbar and profile page

### LINKED ISSUE #567: Support "verified email sessions" users who aren't logged in for navbar and profile page
If a user is redirected from a verified 3rd party partner, they shouldn't be asked to "log in" because we have a session id that validates their identity. 

For the sake of this ticket an "verified email session" is a session id associated with a user who has a verified email address attached to their profile.

For these users:
- the navbar should show the logged in navbar state (without crypto wallet info if non exists)
   -  "logging out" would just clear their session in this context
- the profile page should work 

PENDING DESIGN:
a CTA at the top of the profile page for users who aren't logged in to encourage them to "connect to claim their NFTs"

lucasrmp-cb:
**What changed? Why?**

closes #567 

This adds the possibility of a user being logged in the app without being logged in to thirdweb

**UI changes**

https://github.com/Stand-With-Crypto/swc-web/assets/155585835/9024561d-1ab6-419d-a6a3-3cfdf0bb5b4a

https://github.com/Stand-With-Crypto/swc-web/assets/155585835/013a22f1-f188-4024-8ccc-e512df20789a

**Notes to reviewers**

The preview env is using my local dev db, if someone wants to QA and tweak some values either DM me or take it from Vercel

**How has it been tested?**

- [x] Locally
- [x] Development
- [ ] Unit test
- [ ] Functional test

**Change management**
type=nonroutine 
risk=high 
impact=sev4


src/actions/actionCreateUserActionNFTMint.ts
```diff
@@ -8,13 +8,13 @@ import { BigNumber } from 'ethers'
 import { nativeEnum, object, z } from 'zod'
 
 import { getClientUser } from '@/clientModels/clientUser/clientUser'
+import { appRouterGetAuthUser } from '@/utils/server/authentication/appRouterGetAuthUser'
```
travisbloom-cb, line 11: I think for this one we actually want to leave it

travisbloom-cb (changes requested):
- Can we record a video showcasing the flow for an unauth user with a session id but no crypto address?
- I assume we'll need to modify the ux of the navbar dropdown here? i think it assumes we only show email if its an embeddedwallet user

src/utils/server/authentication/appRouterGetAuthUser.ts
```diff
@@ -0,0 +1,54 @@
...
+    },
+  })
+  if (!user) {
+    return null
+  }
+
+  const cryptoAddress = user.primaryUserCryptoAddress
+  return {
+    userId: user.id,
+    address: cryptoAddress ? parseThirdwebAddress(cryptoAddress.cryptoAddress) : '',
```
travisbloom-cb, line 52: Lets make this nullable, so that we can catch any instances that assume this will always exist (now it wont)

travisbloom-cb (approved):

PR #623, opened 2024-03-08
[trivial] Add e2e test for voter registration flow

kl-cb:
**What changed? Why?**
Add E2E test for voter registration flow

**UI changes**
None

**PlanetScale Deploy Request**

**Notes to reviewers**

**How has it been tested?**

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

**Change management**
type=routine 
risk=low 
impact=sev5


travisbloom-cb (changes requested):
To prevent us from spending gas every time we trigger a e2e test, we prob want to mock out https://github.com/Stand-With-Crypto/swc-web/blob/main/src/inngest/functions/airdropNFT.ts#L32 in cypress.

See [this flag](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/utils/shared/executionEnvironment.ts#L9)

Issue #600, opened 2024-03-06
Programmatically detect when transaction fees are high on Base


travisbloom-cb:
_EDIT 2024-03-15_:

there are two parts for this ticket:
- we should create a cron job that will continue to keep airdropping as long as there are user actions that have not received NFTS.
  - however, we need to space the airdrops apart time-wise so that we do not spike the gas fees because we are airdropping so many at once.
  - similarly, prior to airdropping, we should the check the gas fees. if the gas fees exceeds a threshold, the cron job should stop to let the fees go down, and we will wait until the cron job runs again (and hopefully the gas fees go down).
- we should update all action flows that involve airdropping to check the gas fee.
  - if gas fee exceeds threshold, then just do not do anything and let the cron job handle the airdrop.
  - else, we can airdrop right at the moment.

----------

~~Cron job that checks how much a mint would cost (TW has this via their RPC client). If it exceeds some threshold (env var) and it is currently toggled on, trigger a slack alert and set the toggle (upstash redis).~~

~~When it goes back down and we want to toggle off, similarly update upstash and trigger slack alert~~

PR #631, opened 2024-03-09
[bug] Fix "NFT mint" action transaction fee estimation

benson-budiman-cb:
**What changed? Why?**

Relates to #600 - discovered bug when trying to scope out feature

This PR fixes the "NFT mint" action transaction fee estimation by using `estimateGasCost()` on a prepared transaction instead of using the gas price of the Web3 provider. Why? Because the gas _price_ of the provider (_e.g._ Base network) is _not_ the same as the gas _fee_ for a transaction, thus we were displaying the wrong "fee".

**UI changes**

_Web (Coinbase Wallet seems to rounds the fees, but the actual transaction will not have rounded fees)_:

https://github.com/Stand-With-Crypto/swc-web/assets/135282747/c3ce64f1-2330-42ba-a12a-3090d0291f35

_Mobile web (Coinbase Wallet seems to round the fees)_:

https://github.com/Stand-With-Crypto/swc-web/assets/135282747/f47cf1b8-ad53-4e38-bfb9-4aa867675dc6

**Notes to reviewers**

This PR bumps the `@thirdweb-dev/react` package to the latest revision.

**How has it been tested?**

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

**Change management**
type=routine 
risk=low 
impact=sev5


src/components/app/userActionFormNFTMint/useCheckoutController.ts
```diff
@@ -34,7 +37,7 @@ export function useCheckoutController({
     }
 
     const mintFee = mintUnitFee.mul(quantity)
-    const gasFee = gasUnitFee.mul(quantity)
+    const gasFee = gasUnitFee // Do not multiply by `quantity`.
```
lucasrmp-cb, line 40: I think you should change the variable name out of `gasUnitFee` instead of the comment, what do u think?

benson-budiman-cb: @lucasrmp-cb ah you right

lucasrmp-cb (approved):

PR #680, opened 2024-03-15
feat: add partners page

lucasrmp-cb:
**What changed? Why?**

This creates a partners page and removes the partners section from the contribute page

**UI changes**

image

image

**How has it been tested?**

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

**Change management**
type=routine 
risk=low 
impact=sev5


benson-budiman-cb (commented):
@lucasrmp-cb just one design question, but I can approve otherwise.

Is it intentional that, from the `contribute` page, the `Become an industry partner` just does not lead to the `partners` page? To me, it's a little weird that we have two buttons not very distant from each other that lead to the same form, but maybe this design pattern is preferable.

lucasrmp-cb:
> Become an industry partner

@benson-budiman-cb Yeah thats by design, both buttons have the same text and action

PR #694, opened 2024-03-19
Handle undefined result from action

ydruffin-cb:
closes https://github.com/Stand-With-Crypto/swc-web/issues/670
fixes [PROD-SWC-WEB-18Q](https://stand-with-crypto.sentry.io/issues/5029315222/?referrer=github_integration)

**What changed? Why?**
I added a check of the result of an action in withServerActionMiddleware function and throw an error if the result is undefined. This is to avoid reading a field of an undefined result on the client side.

**Notes to reviewers**
I don't think I fully understand the promises in javascript. So I an not fully confident of my solution here. 
I was able to test it on local, but not sure how I can test this on Preview

**How has it been tested?**

- [x] Locally
- [ ] Vercel Preview Branch

**Change management**
type=routine 
risk=low 
impact=sev5

### DIFF
--- a/src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
+++ b/src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
@@ -76,7 +76,7 @@ export function SuggestedScript({
         },
         payload =>
           actionCreateUserActionCallCongressperson(payload).then(actionResult => {
-            if (actionResult.user) {
+            if (actionResult?.user) {
               identifyUserOnClient(actionResult.user)
             }
             return actionResult
@@ -89,6 +89,7 @@ export function SuggestedScript({
         goToSection(SectionNames.SUCCESS_MESSAGE)
       } else {
         setCallingState('error')
+        toastGenericError()
       }
     },
     [addressSchema, dtsiPerson.slug, goToSection, router],
--- a/src/utils/server/withServerActionMiddleware.ts
+++ b/src/utils/server/withServerActionMiddleware.ts
@@ -1,5 +1,6 @@
 import * as Sentry from '@sentry/nextjs'
 import { headers } from 'next/headers'
+import { getLogger } from '@/utils/shared/logger'
 
 // sentry only allows form data to be passed but we want to support json objects as well so we need to make it form data
 const convertArgsToFormData = <T extends any[]>(args: T) => {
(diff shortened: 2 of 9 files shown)


src/utils/server/withServerActionMiddleware.ts
```diff
@@ -30,7 +34,20 @@ export function withServerActionMiddleware<T extends (...args: any) => any>(
...
+          .then((result: any) => {
+            if (result === undefined) {
+              throw new Error('Undefined result')
+            }
+          })
+          .catch(function (e: any) {
+            return logger.error(e)
+          })
+
+        return promise
```
lucasrmp-cb, line 49: Few things to note here:
1. Putting a `.catch` suppresses all internal server errors the actions might want to throw, you are logging and then returning `undefined` for the request 
2. Since you're suppressing errors (including the one you are throwing in line 42) this request will continue to respond with undefined. If the idea is to fail the request if the result is `undefined` removing the `.catch` will work properly
3. If you wanna log the error anyways, just throw the error inside the `.catch` scope
```typescriptreact
.catch(error => {
  logger.error(error)
  throw error
})
```
4. On the `.catch` syntax, we shouldn't use anonymous functions with the `function` keyword (they should be arrow functions) and you don't need to specify a type for the error.
5. You can do all of this inline, `.then` and `.catch` return the promise itself, you can do `return action(...).then(...).catch(...)` normally

lucasrmp-cb (changes requested):

src/utils/server/withServerActionMiddleware.ts
ydruffin-cb: yes, this is why I found my solution weird, but without the catch, I receive an  error that I am throwing an error on an async function without a catch and got unhandled error in Sentry. So in the end I remove one error by replacing by another one... 😕
For the 4th point, I end up using an anonymous function because my IDE keep complaining about that I didn't specify an error type 🤷‍♂️

Anyway I am going to try couple of your suggestion

lucasrmp-cb (approved):

src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
```diff
@@ -89,6 +89,7 @@ export function SuggestedScript({
         goToSection(SectionNames.SUCCESS_MESSAGE)
       } else {
         setCallingState('error')
+        toastGenericError()
```
travisbloom-cb, line 92: dont we show an error with setCallingState('error')

travisbloom-cb (changes requested):

src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
ydruffin-cb: right

travisbloom-cb (approved):

PR #701, opened 2024-03-21
[feature] reenable conditional live airdrop and return proper response for NFT backfill cron job

### LINKED ISSUE #600: Programmatically detect when transaction fees are high on Base

### LINKED ISSUE #704: trigger analytics when we mint NFTs
Ensure we add a property "Cost USD" so we can aggregate by this metric.

benson-budiman-cb:
Closes #600 
Closes #704 

**What changed? Why?**

This PR reenables the conditional live airdrop (where the condition is that the current estimated transaction fee is under a defined threshold). Why? Because ideally we want the user to receive their NFT as soon as possible, but we also do not want to spend too much ETH on airdrop transaction fees. Thus, we perform the live airdrop if transaction fees are low - otherwise, we will skip the live airdrop and let the NFT backfill cron job airdrop the NFT at a later point.

This PR also returns a proper response from the NFT backfill cron job that includes the stop reason and the remaining number of NFT user actions in the backfill backlog. Why? So we can easily audit the result of any run and to see the trend of backlogged user actions.

**UI changes**

No UI changes.

**PlanetScale Deploy Request**

No PlanetScale schema changes.

**Notes to reviewers**

- `ignoreTurnOffNFTMintFlag` has been renamed to `skipTransactionFeeCheck`
- The `AIRDROP_NFT_ETH_TRANSACTION_FEE_THRESHOLD` constant has been moved to a shared folder
- A Mixpanel analytic event is sent when a NFT has been successfully airdropped - the event includes the NFT slug and the transaction fee in USD
- Changed another analytic event name to match conventions

**How has it been tested?**

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

image

image

**Change management**
type=routine 
risk=low 
impact=sev5

### DIFF
--- a/src/app/api/internal/commerce-webhook/route.ts
+++ b/src/app/api/internal/commerce-webhook/route.ts
@@ -62,7 +62,7 @@ export async function POST(request: NextRequest) {
     const pricingValues = extractPricingValues(body)
 
     // Tracking the payment event regardless of type - the donation action is not created here.
-    analytics.track('Coinbase Commerce webhook event received', {
+    analytics.track('Coinbase Commerce Webhook Event Received', {
       creationMethod: 'On Site',
       paymentExpire: body.event.data.expires_at,
       paymentId: body.id,
--- a/src/app/api/internal/commerce-webhook/route.ts
+++ b/src/app/api/internal/commerce-webhook/route.ts
@@ -62,7 +62,7 @@ export async function POST(request: NextRequest) {
     const pricingValues = extractPricingValues(body)
 
     // Tracking the payment event regardless of type - the donation action is not created here.
-    analytics.track('Coinbase Commerce Webhook Event Received', {
+    analytics.track('Coinbase Commerce webhook event received', {
       creationMethod: 'On Site',
       paymentExpire: body.event.data.expires_at,
       paymentId: body.id,
--- a/src/inngest/functions/airdropNFT.ts
+++ b/src/inngest/functions/airdropNFT.ts
@@ -28,6 +28,8 @@ const AIRDROP_NFT_RETRY = 2
 
 const logger = getLogger('airdropNFTWithInngest')
 
+const WEI_TO_ETH_UNIT = 1e-18
+
 export const airdropNFTWithInngest = inngest.createFunction(
   {
     id: AIRDROP_NFT_INNGEST_FUNCTION_ID,
@@ -61,7 +63,7 @@ export const airdropNFTWithInngest = inngest.createFunction(
       mintStatus = transactionStatus.status
       transactionHash = transactionStatus.transactionHash
       transactionFee = new Decimal(
-        Number(transactionStatus.gasLimit) * Number(transactionStatus.gasPrice),
+        Number(transactionStatus.gasLimit) * +Number(transactionStatus.gasPrice) * WEI_TO_ETH_UNIT, // Gas price is in wei, so we need to convert it to ETH.
       )
       attempt += 1
     }
--- a/src/inngest/functions/airdropNFT.ts
+++ b/src/inngest/functions/airdropNFT.ts
@@ -1,4 +1,5 @@
-import { NFTMintStatus } from '@prisma/client'
+import { NFTCurrency, NFTMintStatus } from '@prisma/client'
+import { Decimal } from '@prisma/client/runtime/library'
 import * as Sentry from '@sentry/nextjs'
 import { NonRetriableError } from 'inngest'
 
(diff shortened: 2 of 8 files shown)


src/inngest/functions/airdropNFT.ts
```diff
@@ -62,22 +77,59 @@ export const airdropNFTWithInngest = inngest.createFunction(
...
     if (status === 'errored' || status === 'cancelled') {
       throw new NonRetriableError(
         `airdrop NFT transaction ${transactionHash!} failed with status ${status}`,
       )
     }
+
+    if (status === 'mined') {
+      await step.run('emit-analytics-event', async () => {
+        // All flows to this Inngest function should have passed in the user's primary crypto address.
+        const user = await prismaClient.user.findFirst({
```
travisbloom-cb, line 97: cant we just get the userid passed in to this fn?

src/inngest/functions/airdropNFT.ts
```diff
@@ -62,22 +77,59 @@ export const airdropNFTWithInngest = inngest.createFunction(
...
+          .then(res => {
+            return res?.data.amount ? res?.data.amount : new Decimal(0)
+          })
+          .catch(e => {
+            logger.error(e)
+            return new Decimal(0)
+          })
+        analytics.track('NFT successfully airdropped', {
+          nftSlug: payload.nftSlug,
+          transactionFeeUSD: transactionFee.mul(ratio).toNumber(),
```
travisbloom-cb, line 130: maybe we include the gas value as well as a property

src/inngest/functions/airdropNFT.ts
```diff
@@ -62,22 +77,59 @@ export const airdropNFTWithInngest = inngest.createFunction(
...
+        const ratio = await getCryptoToFiatConversion(NFTCurrency.ETH)
+          .then(res => {
+            return res?.data.amount ? res?.data.amount : new Decimal(0)
+          })
+          .catch(e => {
+            logger.error(e)
+            return new Decimal(0)
+          })
+        analytics.track('NFT successfully airdropped', {
+          nftSlug: payload.nftSlug,
```
travisbloom-cb, line 129: https://github.com/Stand-With-Crypto/swc-web/blob/638-location-specific-pages/docs/Coding%20Conventions.md#L82

src/inngest/functions/backfillNFTCronJob/index.ts
```diff
@@ -155,5 +154,29 @@ export const backfillNFTInngestCronJob = inngest.createFunction(
         BACKFILL_NFT_INNGEST_CRON_JOB_AIRDROP_SLEEP_INTERVAL,
       )
     }
+
+    // Fetch the remaining user actions that are in the backlog.
+    const userActionsRemaining = await step.run('script.get-user-actions-remaining', async () => {
+      const actions = await prismaClient.userAction.findMany({
```
travisbloom-cb, line 160: can prob optimize this with https://www.prisma.io/docs/orm/prisma-client/queries/aggregation-grouping-summarizing

travisbloom-cb (changes requested):

src/inngest/functions/airdropNFT.ts
benson-budiman-cb: yes we can 🔨

src/app/api/internal/commerce-webhook/route.ts
```diff
@@ -62,7 +62,7 @@ export async function POST(request: NextRequest) {
     const pricingValues = extractPricingValues(body)
 
     // Tracking the payment event regardless of type - the donation action is not created here.
-    analytics.track('Coinbase Commerce webhook event received', {
+    analytics.track('Coinbase Commerce Webhook Event Received', {
```
travisbloom-cb, line 65: dont like to change these once they're already live, because then there's multiple events floating in mixpanel and it breaks historical trends

src/inngest/functions/airdropNFT.ts
```diff
@@ -93,23 +95,17 @@ export const airdropNFTWithInngest = inngest.createFunction(
...
-        // All flows to this Inngest function should have passed in the user's primary crypto address.
         const user = await prismaClient.user.findFirst({
           where: {
-            primaryUserCryptoAddress: {
-              cryptoAddress: payload.recipientWalletAddress,
-            },
+            id: payload.userId,
           },
         })
         if (!user) {
```
travisbloom-cb, line 103: i would just use findFirstOrThrow. If we pass in an nonexistant user id this deserves to break

src/inngest/functions/airdropNFT.ts
travisbloom-cb: property names as well :)

travisbloom-cb (changes requested):

src/app/api/internal/commerce-webhook/route.ts
benson-budiman-cb: gotcha, will revert

travisbloom-cb (approved):

PR #729, opened 2024-03-29
[feature] update aggregation calculations to use `totalDonationAmountUsd` and implement auditing cron job

### LINKED ISSUE #660: Denormalize total donations per user and deprecate expensive query usage
We have all sorts of performance issues associated with [this query](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/data/aggregations/getSumDonationsByUser.ts#L33). This ticket includes the work needed to add a new `totalDonationAmountUsd` to the user table, and plug in the logic needed to backfill/keep up to date/use.

There are multiple PRs that we'll need:

- One to add the new db column
- One that proactively updates it in the actions that impact it moving forward (donation webhook and donation mint action) and updates the merge logic appropriately.
- One to add a cron job that audits all the fields for all the users and performs any necessary sums - simply iterate through each user, and for each user, get their NFT mints and donations, then sum that up (or maybe combine with above)
- One to update the current getSumDonationsByUser logic to use this new field instead

benson-budiman-cb:
closes #660 

## What changed? Why?

This PR updates the aggregation calculates to use `totalDonationAmountUsd` appropriately. Why? Because those fields should have been properly backfilled now, and we can make use of that field to write much more performant queries - this new query will reduce the data aggregation time from minutes to ~1 second. 🚀 

This PR also implements an auditing cron job to ensure that each user's `totalDonationAmountUsd` is as expected. This cron job is _super_ similar to the backfill script. However, instead of updating the database, the cron job will emit a Sentry message.

## UI changes

Same UI, but here are some comparison screenshots between using the new query against the current aggregation query on production

Any discrepancies in due to the new query using slightly old production information.

Screenshot 2024-03-29 at 12 52 40 AM

Screenshot 2024-03-29 at 12 52 57 AM

Screenshot 2024-03-29 at 12 57 32 AM

image

## PlanetScale Deploy Request

No PlanetScale schema changes.

## Notes to reviewers

I have already ran the backfill script in production, so all users have a total USD donation amount (if applicable).

## How has it been tested?

My Vercel preview branch is pointing to a database with production information. Normally, the old aggregation would have timed out while generating the static pages, but now the preview branch builds with ease. 🚀 

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Change management

type=routine 
risk=low 
impact=sev5


travisbloom-cb (approved):
its so beautiful

PR #754, opened 2024-04-08
Changes headings font from Raleway to Satoshi

### LINKED ISSUE #747: Update primary font from Raleway to Satoshi
Our primary font used for headers and display titles should be changed from Raleway to Satoshi.

Link to font: https://www.fontshare.com/fonts/satoshi

Link to Figma: https://www.figma.com/file/ZGYKkf5yHXoS2jadfbB4NM/SWC-Design-System?type=design&node-id=5-2&mode=design&t=MJ0eFkjrletXS9Gi-4

paulovictor-cb:
closes #747 

## What changed? Why?

The headings' font was updated from Raleway to Satoshi font to match swc design system. SWC design system font showcase is available [here](https://www.figma.com/file/ZGYKkf5yHXoS2jadfbB4NM/SWC-Design-System?type=design&node-id=5-2&mode=design&t=MJ0eFkjrletXS9Gi-4)

### Web

| Old 👴         | New 👶         |
| -------------- | -------------- |
|image  | image|

## Notes to reviewers
I had to use relative path to load the font because NextJS was complaining about not finding it. I decided to use relative path and store the fonts in public folder instead of creating an assets folder inside app.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Change management

type=routine
risk=low
impact=sev3


paulovictor-cb:
> We'll def want to [scrunch the odometer letter spacing](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/components/ui/animatedNumericOdometer/odometer.module.css#L10) for this change.
> 
> @coinbasemar take a look at preview and let us know what else needs to be tweaked

I dropped the letter-spacing from -1 to -6, but IMO I don't think that this font really suits this case. Maybe using another font for the odometer would be better.

This is how it looks like with -6px of letter spacing.
image

A suggestion:
This is how it looks like using -1px of letter spacing and open-sans font
image

@travisbloom-cb @coinbasemar

travisbloom-cb (approved):

Issue #733, opened 2024-03-31
Dont crash when user has cookies blocked on safari


travisbloom-cb:
[This sentry error](https://stand-with-crypto.sentry.io/issues/5028531279/?environment=production&project=4506490717470720&query=&referrer=issue-stream&statsPeriod=14d&stream_index=2) occurs when a user visits the site on safari with cookies blocked.

Steps to reproduce:
- set "Safari Settings -> Advanced -> Block all cookies"
- load swc in safari

We already [try catch our calls to localstorage](https://github.com/Stand-With-Crypto/swc-web/blob/638-location-specific-pages/src/utils/web/localStorage.ts#L12) for this exact reason, but it looks like some third party libraries like "WalletConnect" call localstorage directly and crash the site.

I think the fix here is we inline a script tag at the top of our app before anything else loads that overrides the global localStorage function with a version that wraps all the calls in try catch. Ugly hack 🤷

benson-budiman-cb:
_2024-04-01-02:38pm UPDATE:_ 

I've tried to overwrite the local storage _functions_ via the [PR](https://github.com/Stand-With-Crypto/swc-web/pull/735) above, but that does not fix the issue - the thing is, _any access to the local storage (even something like `console.log(localStorage)` causes the insecure error, so overwriting the functions does not matter.

If there is an answer to this, then it would involve somehow checking if the browser can access the local storage, but this check would have to be built part of the local storage because it needs to essentially intercept any calls to `localStorage`. Most solutions out there involve a wrapper function/wrapper around the local storage to do this, but we can't do that given that it's a third-party package that is hitting the local storage directly.

PR #758, opened 2024-04-08
dont crash app when browser errors on localstorage

travisbloom-cb:
closes https://github.com/Stand-With-Crypto/swc-web/issues/733

## Notes to reviewers

See issue for QA notes

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Change management

type=routine 
risk=low 
impact=sev5

### DIFF
--- a/src/components/app/overrideGlobalLocalStorage.tsx
+++ b/src/components/app/overrideGlobalLocalStorage.tsx
@@ -1,0 +1,52 @@
+import Script from 'next/script'
+
+// we include this so we can typecheck/pretty format the code that will end up stringified below
+// function referenceCodeThatGetsCommmentedOut() {
+//   try {
+//     localStorage.getItem('doesNotExist')
+//   } catch (e) {
+//     const localStorageNoop = {
+//       length: 0,
+//       clear: () => null,
+//       getItem: () => null,
+//       key: () => null,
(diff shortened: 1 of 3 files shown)


src/components/app/overrideGlobalLocalStorage.tsx
```diff
@@ -0,0 +1,52 @@
...
+          removeItem: () => null,
+          setItem: () => null,
+        }
+        window.localStorage = localStorageNoop
+        Object.defineProperty(window, 'localStorage', {
+          value: localStorageNoop,
+          configurable: true,
+          enumerable: true,
+          writable: false,
+        })
```
benson-budiman-cb, line 47: 😮 I am surprised that this works! I do remember trying something like `localStorage = localStorageNoop` before, but I was still seeing the error...  I suppose the `Object.defineProperty` part is also needed?

benson-budiman-cb (approved):

src/components/app/overrideGlobalLocalStorage.tsx
travisbloom-cb: ya i think that was the key. Included both just for good measure

PR #760, opened 2024-04-09
support a/b testing via mixpanel

travisbloom-cb:
closes https://github.com/Stand-With-Crypto/swc-web/issues/143

## What changed? Why?

See the experiments.ts file for commentary/context.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Change management

type=routine 
risk=low 
impact=sev5

### DIFF
--- a/src/app/[locale]/internal/user-experiments/page.tsx
+++ b/src/app/[locale]/internal/user-experiments/page.tsx
@@ -1,0 +1,59 @@
+'use client'
+
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from '@/components/ui/select'
+import { useHasHydrated } from '@/hooks/useHasHydrated'
+import { useLocalUser } from '@/hooks/useLocalUser'
+import { EXPERIMENTS_CONFIG } from '@/utils/shared/experiments'
+import { setLocalUserPersistedValues } from '@/utils/web/clientLocalUser'
+
+export const dynamic = 'error'
+
+export default function ExperimentsHomepage() {
+  const localUser = useLocalUser()
+  const hasHydrated = useHasHydrated()
+  const experiments = localUser.persisted?.experiments
+  return (
+    <div className="container mx-auto mt-10">
+      <div className="space-y-7 lg:space-y-10">
+        {!hasHydrated ? null : !experiments ? (
+          <p>Users who reject cookies are opted out of experiments</p>
+        ) : (
+          Object.keys(EXPERIMENTS_CONFIG).map(_experiment => {
+            const experiment = _experiment as keyof typeof EXPERIMENTS_CONFIG
+            return (
+              <div key={experiment}>
+                <h2 className="mb-2 text-xl font-bold">Experiment {experiment}</h2>
+                <Select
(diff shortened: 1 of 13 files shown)


src/utils/web/clientLocalUser.ts
```diff
@@ -42,6 +45,8 @@ export const bootstrapLocalUser = () => {
   getLocalUser()
 }
 
+const log = (msg: string) => logger.info(`getLocalUser - ${msg}`)
```
lucasrmp-cb, line 48: np: doesn't `getLogger` already do this

lucasrmp-cb (approved):
The typings for experiments config are quite complex, it's smelling a little bit for me (I wasn't able to make them simpler with the same type safety though).
Besides that LGTM

travisbloom-cb:
> The typings for experiments config are quite complex, it's smelling a little bit for me (I wasn't able to make them simpler with the same type safety though). Besides that LGTM

Ya its pretty knarly, but the resulting types are the strongest guarantees I could add

src/app/[locale]/internal/user-experiments/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+  SelectValue,
+} from '@/components/ui/select'
+import { useHasHydrated } from '@/hooks/useHasHydrated'
+import { useLocalUser } from '@/hooks/useLocalUser'
+import { EXPERIMENTS_CONFIG } from '@/utils/shared/experiments'
+import { setLocalUserPersistedValues } from '@/utils/web/clientLocalUser'
+
+export const dynamic = 'error'
+
+export default function ExperimentsHomepage() {
```
Comment (benson-budiman-cb) [inline src/app/[locale]/internal/user-experiments/page.tsx:17]: This internal website is designed to be used by a user to override their own experiment groups, right? To clarify, is it possible to explicitly override the experiment groups for a user that isn't your own?

paulovictor-cb (approved):
LGTM. A question:

Would it be good to decrease the time that the experiment variant, assigned to a user, is stored? For what I saw, it is stored for a year with the other properties, but imagine the following scenario:

We create an experiment and define, initially, two variants. A week later, we receive the request to add a new variant to this experiment. The users that visited the app during this first week will have their variants defined as one of the first two and won't ever be able to be assigned variant 3.

The problem here that I'm thinking is that if we use the metrics of variants and their related results to define which one is better, is wouldn't be a clear result as variants 1 and 2 would have an advantage over variant 3. 

One solution would be to create a new experiment from scratch using the three variants, or decrease the stored time to try to mitigate this difference for analysis later.

src/app/[locale]/internal/user-experiments/page.tsx
Comment (travisbloom-cb) [reply src/app/[locale]/internal/user-experiments/page.tsx:17]: yup. I know product/people will want to see a specific version of an experiment so thats why I added this

travisbloom-cb:
> We create an experiment and define, initially, two variants. A week later, we receive the request to add a new variant to this experiment. The users that visited the app during this first week will have their variants defined as one of the first two and won't ever be able to be assigned variant 3.

Ya, in this scenario we can either 1) change the datetime we view in mixpanel to the point the variant was added. 2) start a new experiment

PR #791, opened 2024-04-22
Fix politicians page e2e tests

### LINKED ISSUE #568: Fix broken e2e tests for politicians page
https://github.com/Stand-With-Crypto/swc-web/blob/main/cypress/e2e/3-page-politicians.cy.ts#L9

eduardopicolo-cb:
closes #568

## What changed? Why?
Fixed broken e2e test for politicians page.

## Notes to reviewers
Screenshot 2024-04-22 at 13 51 33

The test was breaking because Cypress flags the button as not-actionable due to it being outside the viewport. To fix it, we only need to call `scrollIntoView()`. Another option is to use `click({ force: true })` to skip visibility checks, but that's not ideal.

But theres another problem:

Screenshot 2024-04-22 at 14 11 18

Interacting with the filters too fast may introduce too many rerenders and break the test. One band-aid solution would be to call `cy.wait()` after every filter is applied. However, the ideal solution is to let react-table handle the filters state.

I refactored the way filters are handled to be more in line with the react-table documentation. By doing so, rerenders have been reduced and there's no need for `cy.wait()` calls.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Change management

type=routine 
risk=low 
impact=sev5

### DIFF
--- a/src/components/app/dtsiClientPersonDataTable/dataTable.tsx
+++ b/src/components/app/dtsiClientPersonDataTable/dataTable.tsx
@@ -3,15 +3,14 @@
 import React, { useMemo } from 'react'
 import {
   Column,
-  ColumnDef,
   ColumnFiltersState,
-  FilterFnOption,
   flexRender,
   getCoreRowModel,
   getFilteredRowModel,
   getPaginationRowModel,
   getSortedRowModel,
   SortingState,
+  TableOptions,
   useReactTable,
 } from '@tanstack/react-table'
 import { debounce } from 'lodash-es'
@@ -20,11 +19,10 @@ import { useRouter } from 'next/navigation'
 
 import { Person } from '@/components/app/dtsiClientPersonDataTable/columns'
 import { DataTablePagination } from '@/components/app/dtsiClientPersonDataTable/dataTablePagination'
+import { getPersonDataTableFilterFns } from '@/components/app/dtsiClientPersonDataTable/filters'
 import {
-  filterDataViaGlobalFilters,
   getGlobalFilterDefaults,
   GlobalFilters,
-  IGlobalFilters,
 } from '@/components/app/dtsiClientPersonDataTable/globalFiltersUtils'
 import { Button } from '@/components/ui/button'
 import { InputWithIcons } from '@/components/ui/inputWithIcons'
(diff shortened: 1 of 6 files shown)


src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -79,48 +74,47 @@ export const SortableHeader = <TData, TValue>({
...
     },
     globalFilterFn,
+    ...rest,
   })
 
-  const debouncedSetGlobalFilter = useMemo(
-    () => debounce(table.setGlobalFilter, 300),
-    [table.setGlobalFilter],
-  )
+  const debouncedSetGlobalFilter = useMemo(() => debounce(setGlobalFilter, 300), [setGlobalFilter])
```
travisbloom-cb, line 117: Curious if we still need the debounce? I'm not quite sure why we added in the first place @lucasrmp-cb ?

src/components/app/dtsiClientPersonDataTable/index.tsx
```diff
@@ -16,35 +16,46 @@ import { apiUrls } from '@/utils/shared/urls'
...
+    fallbackData: { people: sortDTSIPersonDataTable(initialData) },
+    keepPreviousData: true,
+  })
   const locale = useLocale()
-  const memoizedColumns = useMemo(() => getDTSIClientPersonDataTableColumns({ locale }), [locale])
+
   const passedData = useMemo(() => {
-    return sortDTSIPersonDataTable(data?.people || initialData)
-  }, [data?.people, initialData])
+    if (!data?.people) return
```
travisbloom-cb, line 49: Is this needed? dont we always have people because of the initial server rendered data?

eduardopicolo-cb: Yes, the data is always there on first render. I think it is possible to let react-table handle the `sortDTSIPersonDataTable()` call and remove the useMemo altogether.

travisbloom-cb: are we gonna do that? Or is it not needed?

travisbloom-cb (approved):

src/components/app/dtsiClientPersonDataTable/index.tsx
eduardopicolo-cb: Not needed. It seems that react-table does not have a `globalSortingFn` anymore, so manually sorting the data is the way.

travisbloom-cb (approved):

PR #807, opened 2024-04-30
Prevents overscroll on deeplinks

### LINKED ISSUE #787: Prevent content below dialog from scrolling in deep links
If you open a deep link to one of our actions (https://www.standwithcrypto.org/action/call for example) on an iOS device, the content on the page below will scroll when you scroll. Its kinda weird/janky. Because we now have fully height modals, we prob [dont even need the homepage content underneath](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/components/app/homepageDialogDeeplinkLayout.tsx#L49). Solution I'd propose is we modify these deep link pages to just be the modal content.

paulovictor-cb:
closes #787

## What changed? Why?

- Even though there is a CSS property called overscroll that should prevent the background scrolling behavior, it didn't work in any scenario that I tested. 

P.S.: I also didn't remove the homepage content from the background because on desktop the user can see what is behind the modal because it is not full height/width

## How has it been tested?

- [X] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/[locale]/topLevelClientLogic.tsx
+++ b/src/app/[locale]/topLevelClientLogic.tsx
@@ -70,6 +70,20 @@ const InitialOrchestration = () => {
   return null
 }
 
+export function DeeplinkPagesLogic() {
+  const pathname = usePathname()
+
+  useEffect(() => {
+    if (pathname?.includes('/action/')) {
+      document.body.style.overflow = 'hidden'
+    } else {
+      document.body.style.overflow = 'scroll'
+    }
+  }, [pathname])
+
+  return null
+}
+
 // This component includes all top level client-side logic
 export function TopLevelClientLogic({
   children,
@@ -104,6 +118,7 @@ export function TopLevelClientLogic({
           <InitialOrchestration />
         </Suspense>
         {children}
+        <DeeplinkPagesLogic />
       </ThirdwebProvider>
     </LocaleContext.Provider>
   )
--- a/src/app/[locale]/topLevelClientLogic.tsx
+++ b/src/app/[locale]/topLevelClientLogic.tsx
@@ -15,6 +15,7 @@ import { usePathname, useSearchParams } from 'next/navigation'
 import { useThirdwebAuthUser } from '@/hooks/useAuthUser'
 import { useDetectWipedDatabaseAndLogOutUser } from '@/hooks/useDetectWipedDatabaseAndLogOutUser'
 import { LocaleContext } from '@/hooks/useLocale'
+import { usePreventOverscroll } from '@/hooks/usePreventOverscroll'
 import { SupportedLocale } from '@/intl/locales'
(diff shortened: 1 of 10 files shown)


src/app/[locale]/topLevelClientLogic.tsx
```diff
@@ -70,6 +70,20 @@ const InitialOrchestration = () => {
...
+    if (pathname?.includes('/action/')) {
+      document.body.style.overflow = 'hidden'
+    } else {
+      document.body.style.overflow = 'scroll'
+    }
+  }, [pathname])
+
+  return null
+}
+
```
Comment (mhamendes) [inline src/app/[locale]/topLevelClientLogic.tsx:86]: You can just create a hook called
`useDeepLinkPagesLogic` and call it inside `TopLevelClientLogic` instead of creating a component that renders nothing just to add this logic.

src/components/app/userActionFormEmailCongressperson/index.tsx
```diff
@@ -130,7 +130,7 @@ export function UserActionFormEmailCongressperson({
   return (
     <Form {...form}>
       <form
-        className="flex h-full max-h-dvh flex-col"
+        className="flex h-full min-h-[calc(100dvh-4px)] flex-col"
```
mhamendes, line 133: Why is this calc needed?

src/app/[locale]/topLevelClientLogic.tsx
Comment (paulovictor-cb) [reply src/app/[locale]/topLevelClientLogic.tsx]: Changed

src/components/app/userActionFormEmailCongressperson/index.tsx
paulovictor-cb: There is a bug on desktop only when the device height is not big enough that was creating a scroll of about 2px. This calc doesn't change de UI but removes the unnecessary scroll

src/app/[locale]/topLevelClientLogic.tsx
```diff
@@ -78,6 +79,8 @@ export function TopLevelClientLogic({
   children: React.ReactNode
   locale: SupportedLocale
 }) {
+  usePreventOverscroll()
```
Comment (travisbloom-cb) [inline src/app/[locale]/topLevelClientLogic.tsx:82]: prob better to have this on the action pages instead having it global and conditionally checking pathname

travisbloom-cb (approved):

src/app/[locale]/topLevelClientLogic.tsx
Comment (paulovictor-cb) [reply src/app/[locale]/topLevelClientLogic.tsx:82]: I added this to the actions page in a previous version, but I found that this top logic existed and tried to use it to prevent code duplication. I'll push a commit changing this to localize this logic to each page to prevent unknown bugs in the future

travisbloom-cb (dismissed):

travisbloom-cb (approved):

PR #820, opened 2024-05-04
Do not auto-focus inputs for actions on mobile

matheusmendes-cb:
closes 
#818 

fixes

## What changed? Why?

- Disabled radix-ui Dialog's default behavior of auto-focusing on its first focusable child when opened.
- Focus on input only on Desktop devices.

## Notes to reviewers

This behavior was impacting different flows, but just one of them was focusing on a input (`UserActionFormEmailCongressperson`).

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/ui/dialog/index.tsx
+++ b/src/components/ui/dialog/index.tsx
@@ -12,6 +12,7 @@ import {
   dialogFooterCTAStyles,
   dialogOverlayStyles,
 } from '@/components/ui/dialog/styles'
+import { useIsMobile } from '@/hooks/useIsMobile'
 import { cn } from '@/utils/web/cn'
 import { PrimitiveComponentAnalytics } from '@/utils/web/primitiveComponentAnalytics'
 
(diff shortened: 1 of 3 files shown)


src/components/ui/dialog/index.tsx
```diff
@@ -52,22 +52,28 @@ interface DialogContentProps
...
+>(
+  (
+    { className, children, padding = true, onOpenAutoFocus, closeClassName = '', ...props },
+    ref,
+  ) => (
+    <DialogPortal>
+      <DialogOverlay />
+      <DialogPrimitive.Content
+        className={cn(dialogContentStyles, padding && dialogContentPaddingStyles, className)}
+        onOpenAutoFocus={onOpenAutoFocus ? e => onOpenAutoFocus(e) : e => e.preventDefault()}
```
travisbloom-cb, line 64: I actually think we want to leave this? unless it solves a UX issue we were seeing. The change above makes sense.

travisbloom-cb (changes requested):

travisbloom-cb (approved):

src/components/ui/dialog/index.tsx
matheusmendes-cb: Changed it to only execute the `onOpenAutoFocus` prop on desktop devices or explicitly specified.

travisbloom-cb (approved):

Issue #727, opened 2024-03-28
Improved SWC Sign Up Flow


coinbasemar:
We should update our sign up flow with the following:

- Phone number, first name, last name, address is required to sign up
- Show the "Join" GIF as an incentive to sign up
- Show suggested next actions after signing up

Designs: https://www.figma.com/file/r9ilZQ4QMRpjefdQFq93Az/SWC-SMS?type=design&node-id=1-2947&mode=design

travisbloom-cb:
eng notes: 
- this is designed with full page modals, which is out of scope for this ticket
- We need to defer giving people an NFT for creating an account until they fill in their profile info
- We should require all the profile info to submit the form now

travisbloom-cb:
This is blocked by #725

travisbloom-cb:
We should also structure this as an A/B test, with a control and variant for the new flow. See [experiments](https://github.com/Stand-With-Crypto/swc-web/blob/639-show-whether-a-position-is-for-or-against-crypto/src/utils/shared/experiments.ts#L29)

PR #829, opened 2024-05-07
Improved SWC Sign Up Flow

matheusmendes-cb:
closes  #727 
fixes 

## What changed? Why?

We are going to start an experiment in the signup flow:
- 50% of the users will get the NFT as soon as they login for the first time;
- 50% of the users are only going to get the NFT when they finish updating their profile with all fields required.

This PR creates the new `UpdateUserProfileForm` UI, implements a generic component to handle AB Testing with different UIs and implements the AB test logic to `triggerPostLoginUserActionSteps` and `actionUpdateUserProfile`.

## Notes to reviewers

I had to create a new schema called `zodUpdateUserProfileFormFieldsRequired` to make sure everything will be really required and we are not accepting empty strings.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/actions/actionUpdateUserProfile.ts
```diff
@@ -218,3 +226,35 @@ async function handleCapitolCanaryAdvocateUpsert(
...
+      campaignName: UserActionOptInCampaignName.DEFAULT,
+      actionType: UserActionType.OPT_IN,
+      userActionOptIn: {
+        optInType: UserActionOptInType.SWC_SIGN_UP_AS_SUBSCRIBER,
+      },
+    },
+  })
+
+  if (!optInUserAction) {
+    log(`claimNFTAfterUpdateUserProfile: opt in user action don't exist`)
```
travisbloom-cb, line 253: this should trigger a sentry error because it should never happen

src/components/ui/experimentsTesting/index.ts
```diff
@@ -0,0 +1,31 @@
+import { useLocalUser } from '@/hooks/useLocalUser'
+import { Experiments, ExperimentVariant } from '@/utils/shared/experiments'
+
+function useABTesting<K extends Experiments>({ experimentName }: { experimentName: K }) {
```
travisbloom-cb, line 4: np: maybe we call this `useExperimentName` for consistency?

src/components/ui/experimentsTesting/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  variants: ExperimentsComponentVariants<K>[]
+}
+
+export function ExperimentsTesting<K extends Experiments>({
+  experimentName,
+  variants,
+}: ExperimentsComponentProps<K>) {
+  const variant = useABTesting({ experimentName })
+
+  const ComponentToRender = variants.find(v => v.name === variant)?.component
```
travisbloom-cb, line 28: potential bug here: if you exclude a variant.

Ideally we can change the type signature to error if they aren't included, maybe by changing the prop to a dict with all the variants as keys instead of array?

src/components/app/updateUserProfileForm/step1/variant.tsx
```diff
@@ -0,0 +1,245 @@
...
+    emailAddress: user.primaryUserEmailAddress?.emailAddress || '',
+    phoneNumber: user.phoneNumber || '',
+    hasOptedInToMembership: user.hasOptedInToMembership,
+    hasOptedInToSms: true,
+    address: user.address
+      ? {
+          description: user.address.formattedDescription,
+          place_id: user.address.googlePlaceId,
+        }
+      : {
```
travisbloom-cb, line 63: why isnt this null like the contorl?

travisbloom-cb (changes requested):
Great work!

src/components/ui/experimentsTesting/index.ts
matheusmendes-cb: Fixed.

src/components/app/updateUserProfileForm/step1/variant.tsx
matheusmendes-cb: Because address is not optional anymore, so the zod schema does not accept address as null.

src/components/app/updateUserProfileForm/step1/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { ExperimentsTesting } from '@/components/ui/experimentsTesting'
+import { hasCompleteUserProfile } from '@/utils/web/hasCompleteUserProfile'
+
+import { UpdateUserProfileForm as Control } from './control'
+import { UpdateUserProfileForm as Variant } from './variant'
+
+export function UpdateUserProfileFormABTesting(props: Parameters<typeof Control>[0]) {
+  if (hasCompleteUserProfile(props.user)) return <Control {...props} />
```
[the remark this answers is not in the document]
matheusmendes-cb: I think this handles this case, because if the user falls into the `Control` experiment, it will always render the `Control` component, but if the user falls into the `Variant` experiment, it will only have the NFT if he completes his profile, no?

src/actions/actionUpdateUserProfile.ts
matheusmendes-cb: Fixed

src/components/ui/experimentsTesting/index.ts
matheusmendes-cb: Nice catch, fixed.

src/components/app/updateUserProfileForm/step1/index.tsx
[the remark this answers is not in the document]
travisbloom-cb: no, currently we give the NFT as soon as the user signs up, regardless of whether they fill out their profile. Thats the change your making in this PR :(

src/components/app/updateUserProfileForm/step1/variant.tsx
travisbloom-cb: Got it....my gut is that there's a subtle bug here somewhere but if we've sufficiently QAed the address feature is working than maybe it's fine. The form shows a "required" error if you submit w/o adding adding an address? Is the issue that the type definition for default values is wrong if its null? Because its fine of the submit fn expects it not to be null (it should catch and error in that case).

src/components/ui/googlePlacesSelect/index.tsx
```diff
@@ -46,12 +53,13 @@ export const GooglePlacesSelect = React.forwardRef<
...
           className={cn(
-            triggerProps.value || 'text-gray-500',
+            'text-muted-foreground',
+            triggerProps.value && 'text-gray-500',
             'h-auto cursor-pointer whitespace-normal',
             triggerProps.open && 'outline-none ring-2 ring-ring ring-offset-2',
             className,
           )}
-          leftIcon={<MapPin className="h-4 w-4 text-gray-500" />}
+          leftIcon={showIcon ? <MapPin className="h-4 w-4 text-gray-500" /> : undefined}
```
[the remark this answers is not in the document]
travisbloom-cb: Synced w/ Mar, ignore my comment

src/components/app/updateUserProfileForm/step1/index.tsx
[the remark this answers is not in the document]
matheusmendes-cb: Oh yeah, that's right, got it, gonna fix that.

src/components/app/updateUserProfileForm/step1/variant.tsx
matheusmendes-cb: I check this case and it wasn't erroring because the address accepted empty strings so I fixed it and tested both cases. Control component allows the user to submit the form without address set, Variant component does not accept it.

lucasrmp-cb (approved):

lucasrmp-cb:
The github closing keyword should be in the same line as the link to the issue (`fixes #123` instead of `fixes\n#123`, this causes the PR to not be linked to the issue in the `development` section

travisbloom-cb (approved):

travisbloom-cb (dismissed):

PR #831, opened 2024-05-08
Optimize hero with Vercel Blob

### LINKED ISSUE #822: Responsively load optimized gif on homepage hero
On desktop we want to load this higher res gif on the homepage CTA, and on mobile we should load the optimized version. Unfortunately NextImage doesn't work with gifs so we'll need some manual [clientside hooks](https://github.com/Stand-With-Crypto/swc-web/blob/state-page-design-tweaks/src/hooks/useIsDesktop.ts#L3) to accomplish.

Image

eduardopicolo-cb:
closes #822 

## What changed? Why?

Optimized gif and converted to mp4 format, reducing file size from ~7mb to ~300kb with minimal quality loss.
The first frame of the video is used as a fallback while the media loads and also if the client does not support \<video\> tag.

Hosting the media file on [Vercel Blob](https://vercel.com/docs/storage/vercel-blob?utm_source=next-site&utm_medium=docs&utm_campaign=next-website).

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/ui/video.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * `Video` is a component that fetches a video from the Vercel Blob Storage and renders it in a video tag.
+ *
+ * Needs to be wrapped in a `Suspense` component to handle loading states.
+ */
+export async function Video(props: VideoProps) {
+  const { fileName, type = 'video/mp4', className, fallback, ...rest } = props
+
+  const { blobs } = await list({
```
travisbloom-cb, line 23: rather than fetching from blob, just copy the URL and use it directly. Saves a roundtrip for the client

travisbloom-cb (changes requested):

travisbloom-cb (dismissed):
Only minor feedback I have is that the background is gray for split second before it fully loads. If there's a way to set the background to be a purple gradiant similar to the video that would be more seamless, but if thats hard/not feasible LGTM

eduardopicolo-cb:
> Only minor feedback I have is that the background is gray for split second before it fully loads. If there's a way to set the background to be a purple gradiant similar to the video that would be more seamless, but if thats hard/not feasible LGTM

Easily doable

travisbloom-cb (approved):

Issue #837, opened 2024-05-09
If we deep link a user to update their profile and they arent logged in, ensure we redirect to the profile update form after they log in


travisbloom-cb:
Currently we have logic where after a user signs up for the first time, we redirect them to update their profile. We also have a deep link (https://www.standwithcrypto.org/profile?hasOpenUpdateUserProfileForm=true) to send someone directly to the update profile form.

The current feature gap is that if someone previously signed up, and we deep link them to their update user profile form, and they arent logged in, they'll get prompted to login and then wont get redirected to the profile page. We need to ensure that when we link someone to a auth-gated page (like profile) and they arent logged in, we redirect them back after login.

PR #840, opened 2024-05-09
Redirect to the profile update form after they log in on user deep link

matheusmendes-cb:
closes 
#837 

fixes 

## What changed? Why?

To redirect the user back to the profile update form if we deep link a user to update their profile and they arent logged in we have to persist this data somewhere, in this case it was simpler to persist this queryParam on the URL, so now we redirect to the login url persisting the queryParam and once the user is authenticated, we get check if the user has the correct queryParam and if so, we redirect him back to the deep link url.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
```diff
@@ -15,11 +17,15 @@ export default function UserActionOptInSWCDeepLink() {
   const urls = useIntlUrls()
   const router = useRouter()
   const session = useSession()
+  const { queryString } = useQueryParamState({
+    queryParamKey: OPEN_UPDATE_USER_PROFILE_FORM_QUERY_PARAM_KEY,
```
Comment (travisbloom-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:21]: Thoughts on this being generic so you can pass any arbirary redirect path you want here? Like `redirectAfterLogin=/foo/bar?baz=1`

src/types/index.ts
```diff
@@ -6,6 +6,7 @@ export type PageProps<Params = object> = {
   params: Params & {
     locale: SupportedLocale
   }
+  searchParams?: { [key: string]: string | string[] | undefined }
```
travisbloom-cb, line 9: TIL

travisbloom-cb (changes requested):

src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
Comment (travisbloom-cb) [reply src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:21]: You need it to be generic to support the follow up ticket

Comment (matheusmendes-cb) [reply src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:21]: Yes, I realized it after seeing the other ticket. Fixing it right now.

src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
```diff
@@ -15,11 +16,22 @@ export default function UserActionOptInSWCDeepLink() {
...
   const session = useSession()
+  const searchParams = useSearchParams()
+
+  const { destination } = getCallbackDestination({
+    queryString: searchParams?.toString(),
+    defaultDestination: 'profile',
+  })
+
+  const handleRedirectOnLogin = React.useCallback(() => {
+    return router.replace(urls[destination]())
```
Comment (travisbloom-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:27]: can we gracefully fail/trigger sentry here?

travisbloom-cb (changes requested):

src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
Comment (matheusmendes-cb) [reply src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:42]: Done

src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
```diff
@@ -18,14 +19,28 @@ export default function UserActionOptInSWCDeepLink() {
...
   const { destination } = getCallbackDestination({
-    queryString: searchParams?.toString(),
+    queryString,
     defaultDestination: 'profile',
   })
 
   const handleRedirectOnLogin = React.useCallback(() => {
+    // should always have a destination, but if something happens and thats no the case,
+    // we should gracefully fail, redirecting the user to the profile page
+    if (!destination) {
```
Comment (travisbloom-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:32]: the bug im looking to fix is what if there's a destination and its an invalid urls param. We actually dont wanna trigger sentry errors if the user doesnt pass destination

travisbloom-cb (changes requested):

src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
Comment (matheusmendes-cb) [reply src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx]: Fixed!

travisbloom-cb (approved):

PR #843, opened 2024-05-10
Verifies if the informed address is from US and specific enough

### LINKED ISSUE #827: Improvement on handling addresses in the update profile action
For addresses from the US, we want to show an error message informing that the filled address is not specific enough when the get the representative request fails because of the address not being specific enough.

paulovictor-cb:
closes #827 

## What changed? Why?
To prevent US users from setting a not-specific enough address so that we can't find the corresponding dtsi person, we implemented this validation on the update your profile form. The user can continue without informing the address if the update profile form is rendered from inside the profile page, but whenever the user decides to inform any address, it'll be validated to guarantee that it is quality information.

## How has it been tested?

- [X ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/updateUserProfileForm/step1.tsx
+++ b/src/components/app/updateUserProfileForm/step1.tsx
@@ -1,5 +1,5 @@
 'use client'
-import { useRef } from 'react'
+import { useEffect, useRef } from 'react'
 import { useForm } from 'react-hook-form'
 import { zodResolver } from '@hookform/resolvers/zod'
 import * as Sentry from '@sentry/nextjs'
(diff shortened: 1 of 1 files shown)


src/components/app/updateUserProfileForm/step1.tsx
```diff
@@ -70,6 +72,24 @@ export function UpdateUserProfileForm({
...
     defaultValues: defaultValues.current,
   })
+
+  const addressField = form.watch('address')
+
+  const { data: resolvedCongressPersonData, isLoading: isLoadingCongressPersonData } =
+    useCongresspersonData({ address: addressField ?? undefined })
+
+  const shouldDisableNextButton =
+    form.formState.isSubmitting || isLoadingCongressPersonData || !isEmpty(form.formState.errors)
```
travisbloom-cb, line 82: The ux is weird now because the button is disabled even tho the error isnt visible yet (because it hasnt been touched)
image

Why do we need `!isEmpty(form.formState.errors)`

travisbloom-cb (changes requested):

paulovictor-cb:
@travisbloom-cb I think that I'll close this issue and PR for the moment as the changes that we deployed yesterday seems to also enforce that the address is specific enough to find a related congressperson. What do you think?

PR #846, opened 2024-05-10
fix: combobox scrolling top on web

matheusmendes-cb:
closes 

fixes 

Previous PR fixing bug where opening the address with keyboard open on mobile devices was causing the page to crash created another bug where the page is scrolled top on web also. This PR fixes this bug.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/hooks/useGetElementDimensions.ts
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+export function useGetElementDimensions(ref: React.RefObject<HTMLElement>) {
```
travisbloom-cb, line 3: Why the change from `useResizeObserver`?

matheusmendes-cb: It wasn't working, pretty much. When resizing the screen from mobile to web, the dialog on web was getting a width of `0px`. Now it behaves as expected, every time the input is resized, the dialog is resized as well

matheusmendes-cb: Also, the safe useResizeObserver hook is quite more complex, [check it out](https://usehooks-ts.com/react-hook/use-resize-observer). I don't think we need something like that for now.

travisbloom-cb (approved):

PR #860, opened 2024-05-14
feat: adds type validation for google places api results

### LINKED ISSUE #850: Ensure that users addresses are specific enough in email congressperson flow
We're getting a [ton of errors](https://stand-with-crypto.sentry.io/issues/5200762233/?environment=production&project=4506490717470720&query=&referrer=issue-stream&statsPeriod=30d&stream_index=0) from capital canary that email your congressperson flow is not able to find officials in someones area. When you drill into [inngest examples](https://app.inngest.com/env/production/functions/swc-web-capitol-canary.email-rep/logs/01HXS0MCH8CSJ2MD0453WNRP2D) you can see that the user address is not specific enough. The issue is that the google civic API will match against a city level address like "Owensboro, KY, USA" but CC needs a specific address for the email flow to work.

We need to filter our address input down to include only actual user addresses. See this stack overflow thread on options to test out https://stackoverflow.com/questions/15071589/restrict-google-places-autocomplete-to-return-addresses-only

paulovictor-cb:
closes #850 

## What changed? Why?

Added some type validations for google places api results to prevent returning vague addresses that would cause Capitol Canary to fail.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/cypress/e2e/4-action-email-your-congressperson.cy.ts
+++ b/cypress/e2e/4-action-email-your-congressperson.cy.ts
@@ -15,13 +15,6 @@ it('action - email your congressperson', () => {
   cy.contains('Please enter a valid email address')
   cy.contains('Please select a valid address')
 
-  // validate invalid address
-  cy.selectFromComboBox({
-    trigger: cy.get('input[placeholder="Your full address"]'),
-    searchText: 'new york',
-  })
-  cy.contains('Please enter a specific address that includes street-level information')
-
   // validate success
   cy.get('input[placeholder="Your first name"]').type('John')
   cy.get('input[placeholder="Your last name"]').type('Doe')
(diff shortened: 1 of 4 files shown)


cypress/e2e/4-action-email-your-congressperson.cy.ts
paulovictor-cb, line 1: I removed this case because the user won't be able to select any addresses by just typing new york or something generic and vague

cypress/e2e/8-action-call-your-congressperson.cy.ts
paulovictor-cb, line 1: As we have some constraints on what type of results is returned on google places api, I had to change this to be more specific

src/components/ui/googlePlacesSelect/index.tsx
paulovictor-cb, line 1: Added these types based on this doc: https://developers.google.com/maps/documentation/geocoding/requests-geocoding?hl=pt-br#Types

cypress/e2e/8-action-call-your-congressperson.cy.ts
travisbloom-cb: but should the error still show up?

paulovictor-cb: The error is showing up saying that the address is not from the US

travisbloom-cb (approved):

PR #876, opened 2024-05-16
Calculates next action using campaign

### LINKED ISSUE #875: Evaluate next actions taking campaigns for consideration
Currently, next actions, on success screen after a user completes an action, are only calculated based on if X action has been completed before or not ignoring available campaigns for that X action. We should also check if the current campaign was completed or not. For example: we had an email campaign on 2023, but we have a different one for 2024. Because of this, we should present email as a possible next action because the current campaign is different and may not be completed yet.

paulovictor-cb:
closes #875 

## What changed? Why?

This PR changes how we calculate the next action suggestion based on performed user actions. Now we are also checking if the user has completed X action from the current campaign.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/api/identified-user/performed-user-action-types/route.ts
+++ b/src/app/api/identified-user/performed-user-action-types/route.ts
@@ -1,6 +1,6 @@
 import 'server-only'
 
-import { uniq } from 'lodash-es'
+import { uniqBy } from 'lodash-es'
 import { NextResponse } from 'next/server'
 
 import { getMaybeUserAndMethodOfMatchWithMaybeSession } from '@/utils/server/getMaybeUserAndMethodOfMatch'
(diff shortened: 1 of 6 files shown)


src/app/api/identified-user/performed-user-action-types/route.ts
```diff
@@ -12,13 +12,15 @@ async function apiResponseForUserPerformedUserActionTypes() {
...
         userActions: {
-          select: { id: true, actionType: true },
+          select: { id: true, actionType: true, campaignName: true },
         },
       },
     },
   })
 
-  const performedUserActionTypes = uniq(user?.userActions.map(({ actionType }) => actionType))
+  const performedUserActionTypes = uniq(
```
travisbloom-cb, line 21: we need uniqBy here now

travisbloom-cb (changes requested):

src/app/api/identified-user/performed-user-action-types/route.ts
paulovictor-cb: Updated

travisbloom-cb (approved):

PR #878, opened 2024-05-16
remove vercel analytics

travisbloom-cb:
## What changed? Why?
We arent using vercel analytics so removing from our codebase to stop paying for it.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/serverAnalytics/serverAnalytics.ts
```diff
@@ -85,15 +83,6 @@ export function getServerAnalytics(config: ServerAnalyticsConfig) {
       }).catch(onError({ eventName, eventProperties })),
     )
 
-    if (process.env.VERCEL_URL) {
```
paulovictor-cb, line 88: Is this env still required?

paulovictor-cb (approved):

PR #883, opened 2024-05-17
Adds pizza day redirect

paulovictor-cb:
## What changed? Why?

This PR adds a redirect for pizza day deep link

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -253,6 +253,12 @@ const nextConfig = {
         destination: '/races/:paths*',
         permanent: true,
       },
+      // tweet at person campaigns
+      {
+        source: '/pizza',
+        destination: '/action/tweet-at-person/2024_05_22_PIZZA_DAY',
+        permanent: false,
+      },
     ]
   },
 }


travisbloom-cb (approved):

next.config.js
```diff
@@ -253,6 +253,12 @@ const nextConfig = {
         destination: '/races/:paths*',
         permanent: true,
       },
+      // tweet at person campaigns
+      {
+        source: '/pizza',
+        destination: '/action/tweet-at-person/2024_05_22_PIZZA_DAY',
+        permanent: false,
```
lucasrmp-cb, line 260: This should be permanent, no?

travisbloom-cb: na, vanity urls are always temp so they can be reused later on if needed

travisbloom-cb: open to changing but that's what i was assuming

paulovictor-cb: https://nextjs.org/docs/app/api-reference/next-config-js/redirects

permanent true or false - if true will use the 308 status code which instructs clients/search engines to cache the redirect forever, if false will use the 307 status code which is temporary and is not cached.

PR #901, opened 2024-05-22
Minimize Google Civic Api usage

### LINKED ISSUE #899: Audit our usage of Google Civic API to minimize per-user queries
We currently use the [civic api](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/utils/shared/googleCivicInfo.ts#L129) to match users addresses with their congressional district. Given the [quota constraints of the API](https://coinbase.slack.com/archives/C074HPDK067/p1716382288361219), we should audit all the places we use it and see if there's any unnecessary queries/things that can be optimized. 

Example questions:
- are we fetching from the API when users have an incomplete address/no address populated?
- Are we fetching from the API every time a user loads /politicians or one of the /race pages if they have an address? Can we cache this result so we dont need to fetch every time?

matheusmendes-cb:
closes #899 

fixes 

## What changed? Why?

When user does not have an address we were still making the request to the API because we passed an empty string and did not prevent the request from being made in these cases. This PR prevents this from happening and makes more clear that addresses can be `null` or `undefined`.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


travisbloom-cb (approved):
Lets make sure we QA flows with and without the address being autopopulated. If those arent breaking, lgtm

PR #902, opened 2024-05-22
post-FIT21 CTA & remove FIT21 banner

eduardopicolo-cb:
## What changed? Why?

Add post-FIT21 CTA on homepage and remove FIT21 banner.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/navbar/index.tsx
+++ b/src/components/app/navbar/index.tsx
@@ -1,13 +1,11 @@
 'use client'
 
 import { useCallback } from 'react'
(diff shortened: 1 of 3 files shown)


travisbloom-cb (changes requested):
we should be linking to the bill page from that CTA https://www.standwithcrypto.org/bills/hr4763-118-US

We also should reduce spacing between that new CTA and the title

PR #904, opened 2024-05-23
Adds missing await statement

paulovictor-cb:
## What changed? Why?

This PR adds await on createUserActionDonation call. 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/server/coinbaseCommerce/storePaymentRequest.ts
+++ b/src/utils/server/coinbaseCommerce/storePaymentRequest.ts
@@ -86,7 +86,7 @@ export async function storePaymentRequest(payment: CoinbaseCommercePayment) {
       },
     })
     if (user) {
-      return await createUserActionDonation(user, false, payment)
+      return createUserActionDonation(user, false, payment)
     }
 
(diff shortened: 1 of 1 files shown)


src/utils/server/coinbaseCommerce/storePaymentRequest.ts
```diff
@@ -269,7 +269,7 @@ async function createUserActionDonation(
     })
   }
 
-  const donationAction = prismaClient.userAction.create({
+  const donationAction = await prismaClient.userAction.create({
```
travisbloom-cb, line 272: This is the only place we need an await. You can safely return async promises from a fn without awaiting

travisbloom-cb (changes requested):

src/utils/server/coinbaseCommerce/storePaymentRequest.ts
paulovictor-cb: Changed

matheusmendes-cb (approved):

travisbloom-cb (approved):

PR #918, opened 2024-05-29
Update user action success screen

### LINKED ISSUE #678: Update post action success screen
Deprioritize NFT claim action and focus on next user action + SMS opt in

eduardopicolo-cb:
closes #678 

#### Sign-Up/Sign-In Prompt:
After completing an action, users will now be prompted to sign up/in. 

#### SMS Opt-In:
For users who haven't opted in to receive SMS notifications, a form will be displayed to provide their phone number.

#### Next actions:
Show a prioritized list of uncompleted actions.

##  Notes to reviewers
When clearing your browser to test, make sure these Cookies are not removed:
Screenshot 2024-06-05 at 13 58 05
If they are, reload the page to have them set again.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/userActionFormSuccessScreen/index.tsx
```diff
@@ -1,40 +1,89 @@
...
-  const data = useMemo(() => {
-    if (!userData.data || !performedActionsData.data) {
-      return undefined
+  /**
+   * This effect is to avoid having stale actions data when the user
+   * logs in (from the success screen) after performing an action.
+   */
+  useEffect(() => {
+    if (isLoggedIn) {
+      void performedActionsResponse.mutate()
```
lucasrmp-cb, line 36: Sounds like for me that invalidating the cache is the responsibility of the action form when they are created, not the action success screen. Can we change this so when we call the create action api we invalidate or update the cache?

src/components/app/userActionFormSuccessScreen/index.tsx
```diff
@@ -1,40 +1,89 @@
...
-      performedUserActionTypes,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isLoggedIn])
+
+  if (!isLoggedIn || !user) {
+    return <JoinSWC onClose={onClose} />
+  }
+
+  if (!user.phoneNumber || !user.hasOptedInToSms) {
+    if (fullProfileRequest.isLoading) {
```
lucasrmp-cb, line 46: I don't think we should be accessing this directly (more on a comment bellow), we should change this to use`session.isLoading` instead

src/components/app/userActionFormSuccessScreen/index.tsx
```diff
@@ -1,40 +1,89 @@
...
+        }}
+      />
+    )
+  }
+
+  return (
+    <ScrollArea
+      className={cn(
+        dialogContentPaddingStyles,
+        '-mx-6 -mb-6 overflow-auto max-md:-mt-20 md:-mt-14 lg:max-h-[75vh]',
```
lucasrmp-cb, line 67: Same code smell here

src/components/app/userActionFormSuccessScreen/index.tsx
```diff
@@ -1,40 +1,89 @@
...
-  return <UserActionFormSuccessScreenContent {...props} data={data} />
+
+    return (
+      <SMSOptInContent
+        initialValues={{ phoneNumber: user.phoneNumber }}
+        onSuccess={({ phoneNumber }) => {
+          void fullProfileRequest.mutate({
+            user: { ...user, phoneNumber, hasOptedInToSms: true },
+          })
+          void performedActionsResponse.mutate()
```
lucasrmp-cb, line 57: Exactly this, if a client action changes responses from api endpoints that we need the latest version, we should invalidate it's cache

src/hooks/useSession.ts
```diff
@@ -24,6 +24,7 @@ export function useSession() {
     isLoggedIn,
     isLoggedInThirdweb,
     user: fullProfileRequest.data?.user,
+    fullProfileRequest,
```
lucasrmp-cb, line 27: I don't think we should return this, the hook returns session information and the full profile request status I don't see as part of it. Since we're using `useSWR` you could call `useApiResponseForUserFullProfileInfo` again where you need it, it shouldn't trigger another api call. And if the only thing you need is the `mutate` function you can use the global mutate passing the cache key to invalidate https://github.com/Stand-With-Crypto/swc-web/blob/a2dce55a9e2e288c138590b78309582370006a82/src/components/app/authentication/loginDialogWrapper.tsx#L213-L214

src/utils/shared/getUserActionsProgress.ts
```diff
@@ -0,0 +1,45 @@
...
+  [UserActionType.OPT_IN, true],
+  [UserActionType.TWEET, true],
+  [UserActionType.VOTER_REGISTRATION, true],
+  [UserActionType.LIVE_EVENT, false],
+  [UserActionType.TWEET_AT_PERSON, false],
+])
+
+export const USER_ACTIONS_EXCLUDED_FROM_CTA = Array.from(USER_ACTIONS_AVAILABLE_FOR_CTA)
+  .filter(([_, value]) => !value)
+  .map(([key, _]) => key)
```
lucasrmp-cb, line 19: Why all of this? A simple static array wouldn't be enough? Didn't find any reference of `USER_ACTIONS_AVAILABLE_FOR_CTA` to justify it's existence

src/components/app/userActionRowCTA/userActionRowCTAsList.tsx
```diff
@@ -26,22 +30,21 @@ export function UserActionRowCTAsList({
...
-                  ? 'complete'
-                  : 'incomplete'
-            }
-            {...props}
-          />
+        const state = !performedUserActionTypes
+          ? 'unknown'
+          : performedUserActionTypes.includes(actionType)
+            ? 'complete'
+            : 'incomplete'
```
lucasrmp-cb, line 42: np: can we refactor this to not have this nested ternary?

src/components/app/userActionRowCTA/userActionRowCTAsList.tsx
```diff
@@ -8,15 +8,19 @@ import { USER_ACTION_ROW_CTA_INFO } from '@/components/app/userActionRowCTA/cons
 import { cn } from '@/utils/web/cn'
 import { USER_ACTION_TYPE_CTA_PRIORITY_ORDER } from '@/utils/web/userActionUtils'
 
+export interface UserActionRowCTAsListProps {
+  performedUserActionTypes?: UserActionType[]
+  excludeUserActionTypes?: UserActionType[]
+  className?: string
+  render?: (props: React.ComponentPropsWithoutRef<typeof UserActionRowCTA>) => React.ReactNode
```
lucasrmp-cb, line 15: This should be a component, doing it like this you're removing a part of the lifecycle ou of it (see the 2 other comments bellow for the implementation)
```suggestion
  render?: React.ComponentType<React.ComponentPropsWithoutRef<typeof UserActionRowCTA>>
```

src/components/app/userActionRowCTA/userActionRowCTAsList.tsx
```diff
@@ -8,15 +8,19 @@ import { USER_ACTION_ROW_CTA_INFO } from '@/components/app/userActionRowCTA/cons
...
 export function UserActionRowCTAsList({
   performedUserActionTypes,
   excludeUserActionTypes,
   className,
-}: {
-  className?: string
-  performedUserActionTypes?: UserActionType[]
-  excludeUserActionTypes?: UserActionType[]
-}) {
+  render,
```
lucasrmp-cb, line 22:
```suggestion
  render: Render,
```

src/components/app/userActionRowCTA/userActionRowCTAsList.tsx
```diff
@@ -26,22 +30,21 @@ export function UserActionRowCTAsList({
...
-            {...props}
-          />
+        const state = !performedUserActionTypes
+          ? 'unknown'
+          : performedUserActionTypes.includes(actionType)
+            ? 'complete'
+            : 'incomplete'
+
+        return render ? (
+          render({ state, ...props })
```
lucasrmp-cb, line 45:
```suggestion
          <Render state={state} {...props} />
```

src/components/app/userActionRowCTA/index.tsx
```diff
@@ -64,25 +77,35 @@ export const UserActionRowCTAButton = React.forwardRef<
...
+          <div
+            // Show on Mobile
+            className="sm:hidden"
+          >
+            <div className="mb-1 text-base font-bold lg:text-2xl">{shortText}</div>
+            <div className="text-sm text-gray-500 lg:text-xl">{shortSubtext}</div>
+          </div>
+          <div
+            // Show on Desktop
+            className="max-sm:hidden"
```
lucasrmp-cb, line 101: np: In other places we use the mobile first approach for this, we could unify it to the others
```suggestion
            className="hidden md:block"
```

src/components/app/userActionRowCTA/index.tsx
```diff
@@ -64,25 +77,35 @@ export const UserActionRowCTAButton = React.forwardRef<
         disabled={!canBeActionedOn}
         ref={ref}
       >
-        <div className="flex items-center gap-4">
+        <div className="flex flex-1 items-center gap-4">
           {state !== 'hidden' && <div className="flex-shrink-0">{getStateUI()}</div>}
-          <div className="hidden flex-shrink-0 md:block">
+          <div className="relative flex h-[80px] w-[80px] flex-shrink-0 items-center justify-center overflow-hidden rounded-xl bg-black lg:h-[100px] lg:w-[100px]">
```
lucasrmp-cb, line 82: Is `relative` necessary?

src/components/app/userActionFormVoterRegistration/sections/success.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const UserActionFormVoterRegistrationSuccess = () => {
+  return (
+    <UserActionFormSuccessScreenFeedback
+      Image={
+        <div className="relative h-[180px] w-[345px] overflow-hidden rounded-xl">
+          <NextImage
+            alt={I_AM_A_VOTER_NFT_IMAGE.alt}
+            fill
+            objectFit="cover"
+            src="/actionTypeIcons/registerToVote.png"
```
lucasrmp-cb, line 18: Can we use sizes instead of fill here? Better for optimization

src/components/app/userActionFormVoterRegistration/index.tsx
```diff
@@ -6,12 +6,12 @@ import {
   SectionNames,
 } from '@/components/app/userActionFormVoterRegistration/constants'
 import { ClaimNft } from '@/components/app/userActionFormVoterRegistration/sections/claimNft'
+import { UserActionFormVoterRegistrationSuccess } from '@/components/app/userActionFormVoterRegistration/sections/success'
 import { Survey } from '@/components/app/userActionFormVoterRegistration/sections/survey'
 import { VoterRegistrationForm } from '@/components/app/userActionFormVoterRegistration/sections/voterRegistrationForm'
 import { useSections } from '@/hooks/useSections'
-import { NFTSlug } from '@/utils/shared/nft'
+import { getValues } from '@/utils/shared/getEntries'
```
lucasrmp-cb, line 13: I don't think these utils should exists and we shouldn't use them:
1. `Object.values`, `Object.entries` and `Object.keys` already are typesafe
2. The only thing these functions are doing is suppressing errors that we shouldn't want to suppress, this can cause errors impossible to identify bc nothing is being logged
3. Type safety in the reference object will never allow possibly undefined'ish values to be passed to a native call of `Object.values`

src/components/app/userActionFormVoterRegistration/sections/success.tsx
```diff
@@ -0,0 +1,25 @@
...
+const I_AM_A_VOTER_NFT_IMAGE = NFT_CLIENT_METADATA[NFTSlug.I_AM_A_VOTER].image
+
+export const UserActionFormVoterRegistrationSuccess = () => {
+  return (
+    <UserActionFormSuccessScreenFeedback
+      Image={
+        <div className="relative h-[180px] w-[345px] overflow-hidden rounded-xl">
+          <NextImage
+            alt={I_AM_A_VOTER_NFT_IMAGE.alt}
+            fill
```
lucasrmp-cb, line 16: Same question of `fill` against `sizes` here

src/components/app/userActionRowCTA/index.tsx
```diff
@@ -64,25 +77,35 @@ export const UserActionRowCTAButton = React.forwardRef<
...
+              className="object-cover lg:h-[80px] lg:w-[80px]"
+              height={60}
+              sizes="(max-width: 768px) 60px, 80px"
+              width={60}
+              {...image}
             />
           </div>
-          <div>
+          <div
+            // Show on Mobile
```
lucasrmp-cb, line 93: Unnecessary comment

src/components/app/userActionRowCTA/index.tsx
```diff
@@ -64,25 +77,35 @@ export const UserActionRowCTAButton = React.forwardRef<
...
+            className="sm:hidden"
+          >
+            <div className="mb-1 text-base font-bold lg:text-2xl">{shortText}</div>
+            <div className="text-sm text-gray-500 lg:text-xl">{shortSubtext}</div>
+          </div>
+          <div
+            // Show on Desktop
```
lucasrmp-cb, line 100: Unnecessary comment

src/components/app/userActionRowCTA/index.tsx
lucasrmp-cb, line 101: Same thing here, with the idea of mobile first would be better to do `block sm:hidden` here

src/components/app/userActionRowCTA/index.tsx
```diff
@@ -64,25 +77,35 @@ export const UserActionRowCTAButton = React.forwardRef<
...
+            className="max-sm:hidden"
+          >
             <div className="mb-1 text-base font-bold lg:text-2xl">{text}</div>
             <div className="text-sm text-gray-500 lg:text-xl">{subtext}</div>
           </div>
         </div>
         {canBeActionedOn ? (
-          <div>
+          <div className="max-sm:hidden">
```
lucasrmp-cb, line 108: +1

src/components/app/userActionFormSuccessScreen/smsOptIn.tsx
```diff
@@ -0,0 +1,47 @@
+'use client'
```
lucasrmp-cb, line 1: Why is this a client component?

src/components/app/userActionFormSuccessScreen/smsOptIn.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { SMSOptInForm, SMSOptInFormProps } from '@/components/app/smsOptInForm'
+import { UserActionFormSuccessScreenFeedback } from '@/components/app/userActionFormSuccessScreen/UserActionFormSuccessScreenFeedback'
+import { Skeleton } from '@/components/ui/skeleton'
+
+interface SMSOptInContentProps extends Omit<SMSOptInFormProps, 'children'> {}
+
+export function SMSOptInContent(props: SMSOptInContentProps) {
+  return (
+    <div className="mx-auto flex h-full max-w-[60ch] flex-col items-center gap-6 text-center">
```
lucasrmp-cb, line 11: Instead of using literal values for everything in this cases we should try and use built in values for UI consistence, we should revisit this PR and take a look at which of these values are actually necessary to be a one-of

lucasrmp-cb (changes requested):

src/components/app/userActionFormJoinSWC/successDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+      action => action.actionType,
+    )
+
+  return (
+    <Dialog {...dialogProps}>
+      <DialogContent className="max-w-3xl ">
+        <ScrollArea
+          className={cn(
+            dialogContentPaddingStyles,
+            '-mx-6 -mb-6 overflow-auto max-md:-mt-20 md:-mt-14 lg:max-h-[75vh]',
```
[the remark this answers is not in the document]
eduardopicolo-cb: Without negative margin the scrollbar looks like this 
image

Or we could let it overflow the modal:
image

public/actionTypeIcons/call.png
olavoparno, line 1: hey man, why are we deleting the svg files? I think we use at `src/components/app/recentActivityRow/activityAvatar.tsx`, don't we?

I only caught that because I was using the same ones on the map feature, same ones from the homepage actions activity

src/components/app/pageBillDetails/filters.tsx
```diff
@@ -14,10 +14,9 @@ import {
   DTSI_PersonPoliticalAffiliationCategory,
   DTSI_PersonRoleCategory,
 } from '@/data/dtsi/generated'
-import { getValues } from '@/utils/shared/getEntries'
 import { cn } from '@/utils/web/cn'
 
-export const STANCE_OPTIONS = ['All', ...getValues(DTSI_BillPersonRelationshipType)] as const
+export const STANCE_OPTIONS = ['All', ...Object.values(DTSI_BillPersonRelationshipType)] as const
```
olavoparno, line 19: why was getValues deleted? sorry, still getting acquainted to the codebase

olavoparno: oh, Object.values is generic now, is that it?

public/actionTypeIcons/call.png
eduardopicolo-cb: The ones being used in the activity feed comes from `/public/activityFeedIcons`. 
The ones I deleted from `/public/actionTypeIcons` were not being used anywhere.

olavoparno: Right, my bad thank you

src/components/app/userActionFormSuccessScreen/constants.ts
```diff
@@ -0,0 +1,46 @@
...
+  [UserActionType.CALL]: {
+    title: 'You called your senator!',
+    description: DEFAULT_USER_ACTION_FORM_SUCCESS_SCREEN_INFO['WITH_NFT'],
+  },
+  [UserActionType.TWEET]: {
+    title: 'You helped spread the word!',
+    description: DEFAULT_USER_ACTION_FORM_SUCCESS_SCREEN_INFO['WITHOUT_NFT'],
+  },
+  [UserActionType.DONATION]: {
+    title: 'You donated to help keep  crypto in America!',
```
olavoparno, line 38: double space between keep and crypto it seems

src/components/app/pageBillDetails/filters.tsx
eduardopicolo-cb: Per Lucas request https://github.com/Stand-With-Crypto/swc-web/pull/918#discussion_r1629853196.

The util function could suppress errors and the return type is the same as Object.values.

olavoparno (dismissed):

twistershark (dismissed):

twistershark (dismissed):

twistershark (dismissed):

twistershark (approved):

PR #919, opened 2024-05-29
Prevent Outlook error spikes

### LINKED ISSUE #848: Deal with error spikes caused by SWC email campaigns
Every time SWC team sends emails out to advocates we see a spike in error rates on the site. See this [mixpanel view](https://mixpanel.com/s/4HlVE). Looking in sentry to correlate the error pages to actual exceptions yields examples [like this](https://stand-with-crypto.sentry.io/issues/5037833904/?environment=production&project=4506490717470720&query=is%3Aunresolved+timesSeen%3A%3E%3D10&referrer=issue-stream&statsPeriod=30d&stream_index=0). The error message is not helpful, but if you scroll down to the breadcrumbs you see we log the message "antifingerprint not defined yet. will try and handle event after its ready".

Investigating this error leads to [this insight](https://github.com/getsentry/sentry-javascript/issues/9143#issuecomment-2002664796). It appear whats happening is that a percentage of advocates are using outlook, and their outlook app is scanning SWC when the email arrives.

We need to:
- figure out how ignore these kinds of email bot scans in Sentry
- figure out how to ignore these kinds of events in mixpanel

OR

- figure out to get the error to not occur (which seems unlikely)

paulovictor-cb:
closes #848 

## What changed? Why?

This PR tries to add a conditional to don't log the errors caused by outlook bots from showing up in sentry and mixpanel.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -75,7 +75,6 @@ Sentry.init({
     // `Can't find variable: bytecode`,
     'ResizeObserver loop limit exceeded',
     'ResizeObserver loop completed with undelivered notifications',
-    /Non-Error promise rejection captured with value: Object Not Found/,
   ],
   beforeSend: (event, hint) => {
     // prevent local errors from triggering sentry
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -75,6 +75,7 @@ Sentry.init({
     // `Can't find variable: bytecode`,
     'ResizeObserver loop limit exceeded',
     'ResizeObserver loop completed with undelivered notifications',
+    /Non-Error promise rejection captured with value: Object Not Found/,
   ],
   beforeSend: (event, hint) => {
     // prevent local errors from triggering sentry
--- a/src/hooks/useHandlePageError.ts
+++ b/src/hooks/useHandlePageError.ts
@@ -17,14 +17,15 @@ export function useHandlePageError({
     const isIntentionalError = window.location.pathname.includes('debug-sentry')
     logger.info(`${humanReadablePageName} Error Page rendered with:`, error)
     Sentry.captureException(error, { tags: { domain } })
-
-    if (checkIfErrorIsCausedByOutlook(error)) return
-
     const message = isIntentionalError
       ? `Testing Sentry Triggered ${humanReadablePageName} Error Page`
       : `${humanReadablePageName} Error Page Displayed`
     Sentry.captureMessage(message, { fingerprint: [`fingerprint-${message}`] })
-    trackClientAnalytic('Error Page Visible', { Category: humanReadablePageName })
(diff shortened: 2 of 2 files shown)


sentry.client.config.ts
```diff
@@ -75,6 +75,7 @@ Sentry.init({
     // `Can't find variable: bytecode`,
     'ResizeObserver loop limit exceeded',
     'ResizeObserver loop completed with undelivered notifications',
+    /Non-Error promise rejection captured with value: Object Not Found/,
```
travisbloom-cb, line 78: lets not ignore this in code (we can ignore in sentry)

src/hooks/useHandlePageError.ts
```diff
@@ -17,10 +17,35 @@ export function useHandlePageError({
     const isIntentionalError = window.location.pathname.includes('debug-sentry')
     logger.info(`${humanReadablePageName} Error Page rendered with:`, error)
     Sentry.captureException(error, { tags: { domain } })
+
+    if (checkIfErrorIsCausedByOutlook(error)) return
```
travisbloom-cb, line 21: lets add a `Cause: Outlook` property to the event instead of not triggering at all

src/hooks/useHandlePageError.ts
```diff
@@ -17,10 +17,35 @@ export function useHandlePageError({
...
 }
+
+// we are not sure what causes outlook users to trigger an anti-fingerprint error when accessing
+// SWC using the parsed safe link from outlook. This is a fix to prevent errors spikes
+// from showing up in Sentry and Mixpanel when new email campaigns are sent out.
+// You can find more information about this issue here: https://github.com/Stand-With-Crypto/swc-web/issues/848
+const OUTLOOK_BOT_ERROR_MESSAGE =
+  'Non-Error promise rejection captured with value: Object Not Found'
+function checkIfErrorIsCausedByOutlook(error: Error & { digest?: string }) {
+  if (error?.message.includes(OUTLOOK_BOT_ERROR_MESSAGE)) return true
```
travisbloom-cb, line 38: lets be defensive on our .includes here as well on the offchance the error doesnt include

travisbloom-cb (changes requested):

sentry.client.config.ts
paulovictor-cb: Removed

src/hooks/useHandlePageError.ts
paulovictor-cb: updated

src/hooks/useHandlePageError.ts
paulovictor-cb: Np

travisbloom-cb (approved):

Issue #569, opened 2024-03-05
Add e2e tests for mint NFT flow


travisbloom-cb:
https://testing.standwithcrypto.org/action/nft-mint

Open technical questions:
- How can we mock out the Thirdweb auth flow?
-- naive approach would be to have a bunch of conditionals that only work in cypress environment

olavoparno:
We're now using LocalWallets for the cypress environment https://blog.thirdweb.com/guides/how-to-use-local-wallets/

PR #939, opened 2024-06-04
test: add mint nft action cypress

olavoparno:
#569

closes 

fixes 

## What changed? Why?

Introduced LocalWallet for cypresses tests to mock a temporary authentication and added a test flow for the mint nft user action

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/cypress/fixture/mocks.ts
+++ b/cypress/fixture/mocks.ts
@@ -1,0 +1,11 @@
+export const mockWallet = {
+  password: Math.random().toString(36).substring(7),
+}
+
+export const mockRandomUser = {
+  email: `johndoe${Math.random().toString(36).substring(7)}@email.com`,
+  firstName: 'John',
+  lastName: 'Doe',
+  phoneNumber: '1234567890',
+  address: '350 Fifth Avenue New York, NY 10118',
+}
--- a/cypress/fixture/mocks.ts
+++ b/cypress/fixture/mocks.ts
@@ -1,11 +1,20 @@
+import { faker } from '@faker-js/faker'
+
 export const mockWallet = {
-  password: Math.random().toString(36).substring(7),
+  password: faker.internet.password({ length: 6 }),
 }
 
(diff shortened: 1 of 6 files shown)


cypress/fixture/mocks.ts
```diff
@@ -0,0 +1,11 @@
+export const mockWallet = {
+  password: Math.random().toString(36).substring(7),
```
lucasrmp-cb, line 2: Can we use faker here?

cypress/fixture/mocks.ts
```diff
@@ -0,0 +1,11 @@
...
+}
+
+export const mockRandomUser = {
+  email: `johndoe${Math.random().toString(36).substring(7)}@email.com`,
```
lucasrmp-cb, line 6: +1 on the faker question

src/components/app/updateUserProfileForm/step1/control.tsx
```diff
@@ -125,7 +125,7 @@ export function UpdateUserProfileForm({
                 <FormItem>
                   <FormLabel>Email</FormLabel>
                   <FormControl>
-                    <Input placeholder="Your email" {...field} />
+                    <Input data-testid="email-address-input" placeholder="Your email" {...field} />
```
lucasrmp-cb, line 128: Are you using the `data-testid`'s you added in this PR? Not seeing any references

cypress/fixture/mocks.ts
olavoparno: Definitely! Thanks for the suggestion, I wasn't aware we used

src/components/app/updateUserProfileForm/step1/control.tsx
olavoparno: You're right, I'm not, at least for this field. Thanks for noticing

lucasrmp-cb (approved):

PR #941, opened 2024-06-04
Adds cache header for images

### LINKED ISSUE #934: Investigate why NextImage isn't respecting DTSI cache headers
Getting a large increase in bandwidth being sent to DTSI (which isn't great because I pay for it personally). This is being caused by hitting the CDN for politician profile pictures. NextImage should be [caching the image](https://nextjs.org/docs/pages/api-reference/components/image#caching-behavior) on SWC end so the DTSI cdn isn't hit every time but it doesn't appear to be working as expected. This work should:

- verify DTSI CDN images have the appropriate cache headers
- understand why NextImage isn't respecting (or if @travisbloom-cb is misunderstanding the next.js docs)
- Implement fix on SWC end (or suggest DTSI changes if SWC can't resolve)

paulovictor-cb:
closes #934 

## What changed? Why?

This PR adds a new header to control cache following the nextjs documentation for image [optimization](https://nextjs.org/docs/app/api-reference/components/image#caching-behavior)

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -178,6 +178,7 @@ const V1_ACTION_REDIRECTS = ACTION_REDIRECTS.map(({ destination, queryKey, query
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   images: {
+    minimumCacheTTL: 60 * 60 * 24 * 7, // 7 days
     unoptimized: false,
     remotePatterns: [
       {
@@ -192,7 +193,7 @@ const nextConfig = {
     return [
       {
         source: '/(.*)',
-        headers: [...securityHeaders, { key: 'Cache-Control', value: 's-maxage=604800' }], // 7 days
+        headers: securityHeaders,
       },
     ]
   },
--- a/next.config.js
+++ b/next.config.js
@@ -192,7 +192,7 @@ const nextConfig = {
     return [
       {
         source: '/(.*)',
-        headers: securityHeaders,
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -192,7 +192,7 @@ const nextConfig = {
     return [
       {
         source: '/(.*)',
-        headers: securityHeaders,
+        headers: [...securityHeaders, { key: 'Cache-Control', value: 's-maxage=604800' }], // 7 days
```
lucasrmp-cb, line 195: Can't we specify better sources for this cache? This is applying for the whole app

paulovictor-cb: NextJS doc is a bit confusing on this after the app router release. But setting this header would not make any change on app router because the cache is controlled through the revalidate const for each page.

As it didn't work on the first time, I decided to move to minimumCacheTTL as it is scoped for images only and from what I've seen, it looks like the cache was set correctly after the first load 
image

paulovictor-cb: 60 * 60 * 24 * 7 = 604800

lucasrmp-cb (approved):

travisbloom-cb (approved):
If this works, lgtm, but im still a little confused why its not respecting the cache headers that are set on the images already? Did we figure out what was happening there?

PR #949, opened 2024-06-06
New attempt to try to catch outlook error

twistershark:
## What changed? Why?

This PR is a new attempt to try to catch the outlook error using [this](https://github.com/getsentry/sentry-javascript/issues/3440#issuecomment-828834651) suggestion as reference.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/hooks/useHandlePageError.ts
+++ b/src/hooks/useHandlePageError.ts
@@ -34,18 +34,18 @@ export function useHandlePageError({
 }
 
 // we are not sure what causes outlook users to trigger an anti-fingerprint error when accessing
-// SWC using the parsed safe link from outlook. This is a fix to prevent errors spikes
-// from showing up in Sentry and Mixpanel when new email campaigns are sent out.
+// SWC using the parsed safe link from outlook. This is fix to track the error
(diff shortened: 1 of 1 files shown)


src/hooks/useHandlePageError.ts
```diff
@@ -36,36 +37,24 @@ export function useHandlePageError({
...
-  if (
-    error?.digest &&
-    OUTLOOK_BOT_ERROR_MESSAGE.some(errorMsg => error?.digest?.includes(errorMsg))
-  ) {
-    return true
-  }
   if (
-    error?.stack &&
-    OUTLOOK_BOT_ERROR_MESSAGE.some(errorMsg => error?.stack?.includes(errorMsg))
+    error !== undefined &&
```
olavoparno, line 46: we should always assert comparisons in a safer way such as `typeof error !== 'undefined'` what do you think? the same goes to the rest

olavoparno (dismissed):

olavoparno (approved):

olavoparno (approved):

lucasrmp-cb (approved):

Issue #952, opened 2024-06-07
Alert users who come from UK to redirect to uk.standwithcrypto.org


travisbloom-cb:
If a user comes from the UK, we should show a small alert above the hero content on the homepage that says: "Looking for Stand With Crypto UK? Click here" Where "here" is a link to https://uk.standwithcrypto.org" (link will work later this week).

Can use Collapse component to make display animated. Think you might want [navigator.language](https://stackoverflow.com/questions/673905/how-can-i-determine-a-users-locale-within-the-browser) for this but open to other ideas.

PR #956, opened 2024-06-07
feat: add uk website disclaimer with link

olavoparno:
#952

closes 

fixes 

## What changed? Why?

- adds a banner to disclaim the new UK stand with crypto website.
- mobile version the banner becomes clickable in its entirety

image

image

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
+++ b/src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
@@ -28,7 +28,7 @@ export function UKDisclaimerBanner() {
   useEffect(() => {
     if (hasHydrated && showBanner) {
       setIsRendered(true)
-      setTimeout(() => setIsVisible(true), 10) // Delay to ensure rendering before adding transition class
+      setTimeout(() => setIsVisible(true), 10)
     }
   }, [hasHydrated, showBanner])
 
--- a/src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
+++ b/src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
@@ -41,7 +41,7 @@ export function UKDisclaimerBanner() {
         <div className="container flex justify-between">
           <div className="w-full space-y-1 text-sm text-background antialiased max-sm:text-center sm:text-base">
             <p>
-              Looking for Stand With Crypto UK? Click{' '}
+              🇬🇧 Looking for Stand With Crypto UK? Click{' '}
               <strong>
                 <Link href={externalUrls.ukSWCUrl()}>here</Link>
               </strong>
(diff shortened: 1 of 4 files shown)


src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { externalUrls } from '@/utils/shared/urls'
+
+const languages = getNavigatorLanguages()
+
+export function UKDisclaimerBanner() {
+  const hasHydrated = useHasHydrated()
+  const router = useRouter()
+  const isMobile = useIsMobile()
+  const [isVisible, setIsVisible] = useState(false)
+  const [isRendered, setIsRendered] = useState(false)
```
travisbloom-cb, line 18: why do we need this? can't we just use hasHydrated directly?

src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const WrapperContainer = isMobile ? 'button' : 'div'
+
+  const handleWrapperClick = () => {
+    router.push(externalUrls.ukSWCUrl())
```
travisbloom-cb, line 23: You prob want https://github.com/Stand-With-Crypto/swc-web/blob/main/src/components/ui/linkBox/index.tsx#L7 to get the desired full-content click effect

src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  const showBanner = Boolean(languages?.some(language => language.includes('en-GB')))
+
+  useEffect(() => {
+    if (hasHydrated && showBanner) {
+      setIsRendered(true)
+      setTimeout(() => setIsVisible(true), 10)
```
travisbloom-cb, line 31: this is just so it's not as jumpy right?

olavoparno: exactly!

src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
olavoparno: you're probably right, lemme adjust it

src/components/app/ukDisclaimerBanner/ukDisclaimerBanner.tsx
olavoparno: I wanted to have an anchor working normally on desktop and have an overlay on the banner on mobile, that's why I had it like this

olavoparno: I couldn't achieve the same effect :( leaving as it is

travisbloom-cb (approved):

Issue #959, opened 2024-06-11
CNN Email Campaign


coinbasemar:
Context:

We're creating an action on Stand With Crypto to prompt people to email CNN and ask them to include crypto as a presidential debate topic. We can leverage our email flow that we already have in place, but this should be a new action separate from our current email flow. They will both exist concurrently. 

The copy is still WIP (contact @axelmarko-cb for details) but the designs are good to go. This should be treated like any other action card, so it will show up on the homepage as well as the profile page.

Link to Figma: https://www.figma.com/design/MWzmEAYQC1NYZiD4BG0FlI/V2-Fast-Follows?node-id=3811-2010

travisbloom-cb:
We'll want to extend the UserActionEmailRecipient model to include an optional emailAddress string field, we'll need a new campaign for this, and we'll need to modify our logic to allow for multiple email campaigns to be running concurrently.

PR #963, opened 2024-06-12
Changes SWC CTA logic to include campaigns

twistershark:
## What changed? Why?

This PR changes how we get the CTA cards and every logic related to that so that we will be able to have more than one CTA from the same action type but with different campaign. For example: we can have the CTA to email the congressperson and another CTA to email CNN about the presidential debate. The idea is to be able to show two or more CTAs with the same action type, but with different campaigns.

## Notes to reviewers

This PR is related to the task #959 and it will have a fast follow that will have everything related to creating the email CNN flow.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionRowCTA/constants.tsx
+++ b/src/components/app/userActionRowCTA/constants.tsx
@@ -110,5 +110,5 @@ export function getUserActionCTAInfo(actionType: ActiveClientUserActionType, cam
     return USER_ACTION_ROW_CTA_INFO_FROM_CAMPAIGN[campaign]
   }
 
-  return null
+  throw new Error(`No CTA info found for campaign: ${campaign}`)
 }
--- a/src/components/app/userActionRowCTA/constants.tsx
+++ b/src/components/app/userActionRowCTA/constants.tsx
@@ -16,6 +16,7 @@ import { useLocale } from '@/hooks/useLocale'
 import { ActiveClientUserActionType } from '@/utils/shared/activeUserAction'
 import { TOTAL_CRYPTO_ADVOCATE_COUNT_DISPLAY_NAME } from '@/utils/shared/constants'
 import { getIntlUrls } from '@/utils/shared/urls'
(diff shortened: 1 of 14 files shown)


src/components/app/userActionRowCTA/constants.tsx
```diff
@@ -94,3 +95,20 @@ export const USER_ACTION_ROW_CTA_INFO: Record<
...
+  if (!campaign || campaign === USER_ACTION_TO_CAMPAIGN_NAME_DEFAULT_MAP[actionType]) {
+    return USER_ACTION_ROW_CTA_INFO[actionType]
+  }
+
+  if (USER_ACTION_ROW_CTA_INFO_FROM_CAMPAIGN[campaign]) {
+    return USER_ACTION_ROW_CTA_INFO_FROM_CAMPAIGN[campaign]
+  }
+
+  return null
+}
```
lucasrmp-cb, line 114: Wouldn't it be better to return `USER_ACTION_ROW_CTA_INFO[actionType]` instead of `null` in case the campaign doesn't have a cta info? Would simplify usage by not having to deal with null cases

twistershark: My idea on returning null is to prevent any bugs if a dev forget to change on necessary files. For example: If you add  a new CTA to the priority array or to the additional campaigns, but you don't set it here in this file, it won't show anything. If I just returned the default one, we could have scenarios of duplicate CTAs because of this. But I can change this, np. Just tried to prevent unexpected scenarios

lucasrmp-cb: In other places we're relying on type safety for that, I think we should use that approach here too, that way we enforce the engineer doesn't forget anything while not introducing unexpected behaviors in the app.

And actually if that's never expected we could actually throw an error instead of proceeding here

twistershark: Thanks for the suggestion! Updated the PR

lucasrmp-cb (approved):

PR #965, opened 2024-06-13
Creates send CNN email campaign

### LINKED ISSUE #959: CNN Email Campaign

twistershark:
closes #959 

## What changed? Why?

This PR creates a new email campaign to send CNN emails about the presidential debate.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

image
image


src/components/app/userActionFormEmailCNN/skeleton.tsx
```diff
@@ -0,0 +1,94 @@
...
+                <label>Address</label>
+                <Input placeholder="Your address" />
+              </FormItemSkeleton>
+            </div>
+
+            <div className="relative">
+              <FormItemSkeleton>
+                <Textarea
+                  defaultValue={getDefaultText()}
+                  placeholder={`
```
lucasrmp-cb, line 52: Do we need this placeholder? Having a skeleton above it seem to solve it

src/components/app/userActionFormEmailCNN/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+                          {...field}
+                          onBlur={e => {
+                            handleMessageChange()
+                            form.setFocus(
+                              e.relatedTarget?.attributes?.getNamedItem('name')?.value as
+                                | 'lastName'
+                                | 'emailAddress'
+                                | 'address',
+                            )
+                          }}
```
lucasrmp-cb, line 173: Can we abstract and explain the need for this setFocus?

src/components/app/userActionFormEmailCNN/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+      campaignName: UserActionEmailCampaignName.CNN_PRESIDENTIAL_DEBATE_2024,
+      subject: 'Include Crypto In The Debate',
+      message: getEmailMessage({
+        firstName: initialValues?.firstName || user?.firstName,
+        lastName: initialValues?.lastName || user?.lastName,
+      }),
+    },
+  })
+
+  function handleMessageChange() {
```
lucasrmp-cb, line 87: This should be an arrow function

https://github.com/Stand-With-Crypto/swc-web/blob/d1ad729cc2af4277ce87d2c56c9595c0b61afce4/docs/Coding%20Conventions.md#L27

src/components/app/userActionFormEmailCNN/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+    const newMessage = getEmailMessage({ firstName, lastName })
+
+    form.setValue('message', newMessage)
+  }
+
+  React.useEffect(() => {
+    if (isDesktop) {
+      form.setFocus('firstName')
+    }
+  }, [form, isDesktop])
```
lucasrmp-cb, line 99: Setting `autoFocus={isDesktop}` on the firstName input should do the same thing

src/components/app/userActionFormEmailCNN/skeleton.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { ExternalLink, InternalLink } from '@/components/ui/link'
+import { LoadingOverlay } from '@/components/ui/loadingOverlay'
+import { PageSubTitle } from '@/components/ui/pageSubTitle'
+import { PageTitle } from '@/components/ui/pageTitleText'
+import { ScrollArea } from '@/components/ui/scroll-area'
+import { Textarea } from '@/components/ui/textarea'
+import { SupportedLocale } from '@/intl/locales'
+import { getIntlUrls } from '@/utils/shared/urls'
+
+export function UserActionFormEmailCNNSkeleton({ locale }: { locale: SupportedLocale }) {
```
lucasrmp-cb, line 14: Can we disable all of the inputs on this form? This has the same problem as the other email form where we are able to type while in the loading state

lucasrmp-cb (commented):
Great work, just a few details

src/components/app/userActionFormEmailCNN/skeleton.tsx
twistershark: I tried removing the placeholder and I didn't like the result because it showed an empty input with 16 rows. Because this is in the skeleton, I changed this to be the field value instead of placeholder and disabled the field. This way, while it is loading, the user can at least read the prefilled email message. What do you think?

src/actions/actionCreateUserActionEmailCNN.ts
```diff
@@ -0,0 +1,307 @@
...
+    emailSubject: validatedFields.data.subject,
+    emailMessage: validatedFields.data.message,
+  }
+  await inngest.send({
+    name: CAPITOL_CANARY_EMAIL_INNGEST_EVENT_NAME,
+    data: payload,
+  })
+
+  await beforeFinish()
+  logger.info('updated user')
```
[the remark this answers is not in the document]
twistershark: Copied the same action for the email action and only changed what was required. From my understanding, I think that this user logic should be important from the same perspective as the default email flow. 

After all the action logic is complete, it is going to create a new user action and connect it to the user. In this way, the user is updated.

let me know if you think that we should change something related to this

src/components/app/userActionFormEmailCNN/index.tsx
twistershark: Tested and it doesn't work on our scenario. What happens is that on the first dialog load, the form is lazily loaded and the render skeleton then the real form logic causes this to not work as expected. 

But it does work if you close the modal and open again. This time, it will have everything loaded and it will automatically focus on first name, but as it is not the case on the first modal load, I'm keeping this useEffect for now

src/components/app/userActionFormEmailCNN/index.tsx
twistershark: Done

olavoparno (approved):

lucasrmp-cb (approved):

PR #966, opened 2024-06-13
Email message not updating correctly between action modal re-renders

### LINKED ISSUE #960: Fix dtsiSlugs field not updating after closing and opening the email modal
There is an issue that after you open the email flow modal, then you change the address, close the modal and open again, the default email message won't be rendered as it will be showing that you need to inform an address to be able to see the email message

twistershark:
closes #960 

## What changed? Why?

This PR fixes a bug where the dtsiSlugs field value was not being updated correctly when the users closes and then reopens the email flow modal.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionFormEmailCongressperson/index.tsx
+++ b/src/components/app/userActionFormEmailCongressperson/index.tsx
@@ -1,5 +1,5 @@
 'use client'
-import React, { useMemo, useRef, useState } from 'react'
+import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { useForm, useWatch } from 'react-hook-form'
 import { zodResolver } from '@hookform/resolvers/zod'
 import { UserActionType } from '@prisma/client'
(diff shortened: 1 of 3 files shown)


olavoparno (approved):

src/components/app/userActionFormEmailCongressperson/index.tsx
```diff
@@ -183,10 +183,11 @@ export function UserActionFormEmailCongressperson({
     addressField?.description,
   )
 
-  const dtsiPeople =
-    dtsiPeopleFromAddressResponse?.data && 'dtsiPeople' in dtsiPeopleFromAddressResponse.data
+  const dtsiPeople = useMemo(() => {
```
lucasrmp-cb, line 186: Any reason for adding this here?

lucasrmp-cb (approved):

src/components/app/userActionFormEmailCongressperson/index.tsx
twistershark: It's because I'm using this dtsiPeople as a dependency in a useEffect and React was complaining about it.

Issue #929, opened 2024-05-31
Live map of activations


travisbloom-cb:

coinbasemar:
Context:

Link to PRD: https://docs.google.com/document/d/1XHm3YqTh5wdgUrQvq5Fg3hqhbMCHOqjtQFz1Ucnq55M/edit?usp=sharing&resourcekey=0-Nn3qVK2SicPHvMe8FpoHjw

We're creating an interactive map of the United States that shows all sign ups to Stand With Crypto. This will be displayed on monitors at live events and eventually shown on the SWC site.

On the map, dots should appear on a state whenever a new member joins from there. There should be clusters for the 3-5 largest cities where we have users. We should also show a live advocate counter at the top.

Here's a link to the Figma as well as references for the animations:

Figma: https://www.figma.com/design/MWzmEAYQC1NYZiD4BG0FlI/V2-Fast-Follows?node-id=3808-560

Map hover animation: https://interactivegeomaps.com/feature/heatmap/

Advocate counter: https://codepen.io/phillyc/pen/nwMEeZ

cc: @travisbloom-cb

travisbloom-cb:
To reduce scope here:
- Lets use the existing `AnimatedNumericOdometer` component for counter
- Lets see if there's any react friendly maps that we can achieve a similar effect with off the shelf.

PR #968, opened 2024-06-13
feat: add advocates heatmap

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/929

## What changed? Why?

- it adds a normal route under [locale] and a standalone outside of it, without header and footer.
- renders the total number of advocates in the US into each state.
- the map renders dots according to actions being fetch on the go and larger dots for the states with more advocates.

#### routes:
https://swc-web-git-feat-add-heatmap-activations-coinbase-vercel.vercel.app/embedded/map/10
https://swc-web-git-feat-add-heatmap-activations-coinbase-vercel.vercel.app/map

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


lucasrmp-cb:
We're missing the reference to the issue here

olavoparno:
> We're missing the reference to the issue here

thanks, I've just added in the PR description.

src/app/embedded/map/[topStatesLimit]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+    description,
+  }),
+}
+
+export default async function MapPage({ params }: PageProps<{ topStatesLimit: number }>) {
+  const homeDataProps = await getHomepageData()
+  const advocatePerStateDataProps = await getAdvocatesMapData(params.topStatesLimit)
+
+  return (
+    <AdvocatesHeatmap
```
Comment (lucasrmp-cb) [inline src/app/embedded/map/[topStatesLimit]/page.tsx:29]: I have an issue with naming on this page, you have this following component tree
```
MapPage
|    AdvocatesHeatmap
|    |    PageTitle
|    |    PageSubTitle
|    |    AdvocateHeatmapCounter
|    |    RenderMap
```

The problem here is a separation of concerns issue because:
- `AdvocatesHeatmap` is actually the whole page content and we have a naming convention for this page components where we have the prefix "page" followed by the page name, so `pageAdvocatesHeatmap` would be more appropriate
- `RenderMap` is not as generic as the name suggests, a better name for it would be `AdvocatesHeatmap`, since it only renders the US map with required props called `getTotalAdvocatesPerState` and `topStateMarkers` which closely relates it to this specific use case

That said I think a more coherent naming component tree would look something like this:

```
EmbeddedMapPage
|    AdvocatesHeatmapPage (defined in `src/components/app/pageAdvocatesHeatmap`)
|   |    PageTitle
|   |    PageSubTitle
|   |    AdvocateHeatmapCounter
|   |    AdvocatesHeatmap (does the same thing that `RenderMap` currently does)
```

src/components/app/advocatesHeatmap/advocateHeatmapCounter.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from '@/data/pageSpecific/getHomepageData'
+import { useApiHomepageTopLevelMetrics } from '@/hooks/useApiHomepageTopLevelMetrics'
+import { SupportedLocale } from '@/intl/locales'
+import { intlNumberFormat } from '@/utils/web/intlNumberFormat'
+
+const mockDecreaseInValuesOnInitialLoadSoWeCanAnimateIncrease = (
+  initial: Omit<TotalAdvocatesProps, 'locale' | 'sumDonations' | 'countPolicymakerContacts'>,
+): Omit<TotalAdvocatesProps, 'locale' | 'sumDonations' | 'countPolicymakerContacts'> => ({
+  countUsers: {
+    count: roundDownNumberToAnimateIn(initial.countUsers.count, 100),
```
lucasrmp-cb, line 19: np: I think we should think about rounding down a percentage of the count instead of a specific value, we have more than 1MM advocates, animating the last digits only doesn't look all that great to me

src/components/app/advocatesHeatmap/advocateHeatmapCounter.tsx
```diff
@@ -0,0 +1,55 @@
...
+      countUsers: {
+        count: intlNumberFormat(locale).format(values.countUsers.count),
+      },
+    }
+  }, [values, locale])
+
+  return (
+    <div className="flex-shrink-0 rounded-3xl bg-secondary px-0 py-4 text-center md:w-1/2 md:p-6 md:py-6">
+      <AnimatedNumericOdometer
+        className={odometerStyles.odometerSatoshi}
```
lucasrmp-cb, line 48: I don't think the css module should be an api for the references to mess with, does just passing `font-sans` here and applying it with `cn` inside the odometer work?

src/components/app/advocatesHeatmap/advocatesHeatmap.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const advocatesPerState = useApiAdvocateMap(advocatesMapPageData, {
+    topStatesLimit,
+  })
+
+  const markers = useMemo(() => createMarkersFromActions(actions.data), [actions.data])
+  const topStateMarkers = useMemo(
+    () =>
+      createMarkersFromTopAdvocateStates(advocatesPerState.data.advocatesMapData.topAdvocateStates),
+    [advocatesPerState.data.advocatesMapData.topAdvocateStates],
+  )
```
lucasrmp-cb, line 36: Again, isn't there a way to move this processing server side?

lucasrmp-cb (changes requested):
Good job, will need a second pass but there's a few points to take a look already

src/components/app/advocatesHeatmap/advocateHeatmapCounter.tsx
olavoparno: I've added to it because font-family is applied to a span inside an H1, it's not on the top level of the component tree

src/app/api/public/advocates-map/total-advocates-per-state/[topStatesLimit]/route.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function GET(
+  _request: NextRequest,
+  { params }: { params: { topStatesLimit: string } },
+) {
+  const { topStatesLimit } = zodParams.parse(params)
+  const data = await getAdvocatesMapData(topStatesLimit)
+
+  return NextResponse.json(data)
+}
```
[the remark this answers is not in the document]
Comment (olavoparno) [reply src/app/api/public/advocates-map/total-advocates-per-state/[topStatesLimit]/route.ts:24]: we need it as lively as possible to show on monitors at conferences

src/app/embedded/map/[topStatesLimit]/page.tsx
```diff
@@ -25,8 +26,13 @@ export default async function MapPage({ params }: PageProps<{ topStatesLimit: nu
   const homeDataProps = await getHomepageData()
   const advocatePerStateDataProps = await getAdvocatesMapData(params.topStatesLimit)
 
+  // redirects to not found in case of invalid topStatesLimit
```
Comment (lucasrmp-cb) [inline src/app/embedded/map/[topStatesLimit]/page.tsx:29]: I think your code bellow is pretty clear, don't think this is necessary

lucasrmp-cb (commented):
We spoke on DMs about the json file stuff, lmk when we have a decision on that pls

src/components/app/pageAdvocatesHeatmap/advocatesHeatmapPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+    topStatesLimit,
+  })
+
+  const markers = useMemo(() => createMarkersFromActions(actions.data), [actions.data])
+  const topStateMarkers = useMemo(
+    () =>
+      createMarkersFromTopAdvocateStates(advocatesPerState.data.advocatesMapData.topAdvocateStates),
+    [advocatesPerState.data.advocatesMapData.topAdvocateStates],
+  )
+  const totalAdvocatesPerState = useMemo(
```
travisbloom-cb, line 37: we shouldnt use useMemo unless we're actually doing something computationally complex

src/utils/server/metadataUtils.ts
```diff
@@ -47,5 +49,25 @@ export const generateMetadataDetails = ({
       description,
       images: [useImage],
     },
+    metadataBase: new URL('https://www.standwithcrypto.org'),
```
travisbloom-cb, line 52: I dont think we want these spread across all pages, just the two top level layouts. maybe have a `const TOP_LEVEL_METADATA_DETAILS` const that gets exported from this file and used in the two top levels?

travisbloom-cb (approved):

lucasrmp-cb (approved):

PR #972, opened 2024-06-17
Changes DTSI API Webhook revalidation logic

### LINKED ISSUE #811: Trigger revalidatePath for API endpoints doesn't work
When we hit the [DTSI webhook](https://github.com/Stand-With-Crypto/swc-web/blob/re-enable-donations/src/app/api/internal/dtsi-updated-slugs-webhook/route.ts#L47) the dtsiAllPeople is not getting refreshed. the [revalidate flag](https://github.com/Stand-With-Crypto/swc-web/blob/re-enable-donations/src/app/api/public/dtsi/all-people/route.ts#L9) does trigger the refresh on the specified interval.

twistershark:
closes #811 

## What changed? Why?

Didn't find any documentation on NextJS that supports the assumption that we can call revalidatePath for api routes. Because of that, all people DTSI request wasn't updated when the Webhook was called. I checked on NextJS 15 release doc to check if anything like this could be supported in the future, but I didn't find anything.

To be able to handle this cache situation, I decided to use a NextJS feature of revalidating tags for fetch requests. This way, I can revalidate the all people DTSI request based on a tag. It is very similar on how React Query tags work.

Documentation on [revalidateTag](https://nextjs.org/docs/app/api-reference/functions/revalidateTag)

### DIFF
--- a/src/app/api/internal/dtsi-updated-slugs-webhook/route.ts
+++ b/src/app/api/internal/dtsi-updated-slugs-webhook/route.ts
@@ -1,14 +1,15 @@
 import { flatten } from 'lodash-es'
-import { revalidatePath } from 'next/cache'
+import { revalidatePath, revalidateTag } from 'next/cache'
 import { NextRequest, NextResponse } from 'next/server'
 import { z } from 'zod'
 
 import { ORDERED_SUPPORTED_LOCALES } from '@/intl/locales'
 import { getLogger } from '@/utils/shared/logger'
 import { requiredOutsideLocalEnv } from '@/utils/shared/requiredEnv'
-import { apiUrls, getIntlUrls } from '@/utils/shared/urls'
+import { getIntlUrls } from '@/utils/shared/urls'
 
 const logger = getLogger('/api/internal/dtsi-updated-slugs-webhook')
+const DTSI_AllPeopleQueryTag = 'DTSI_AllPeopleQuery'
 
 const zodPayload = z.object({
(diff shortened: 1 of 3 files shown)


src/app/api/internal/dtsi-updated-slugs-webhook/route.ts
```diff
@@ -1,14 +1,15 @@
...
 import { z } from 'zod'
 
 import { ORDERED_SUPPORTED_LOCALES } from '@/intl/locales'
 import { getLogger } from '@/utils/shared/logger'
 import { requiredOutsideLocalEnv } from '@/utils/shared/requiredEnv'
-import { apiUrls, getIntlUrls } from '@/utils/shared/urls'
+import { getIntlUrls } from '@/utils/shared/urls'
 
 const logger = getLogger('/api/internal/dtsi-updated-slugs-webhook')
+const DTSI_AllPeopleQueryTag = 'DTSI_AllPeopleQuery'
```
lucasrmp-cb, line 12: np: Can we change this to constant case, export it and use it in the `fetchDTSI` call?

lucasrmp-cb (dismissed):

src/app/api/internal/dtsi-updated-slugs-webhook/route.ts
twistershark: Updated!

lucasrmp-cb (approved):

PR #994, opened 2024-06-19
Changes google places select results to be more strict in US

### LINKED ISSUE #947: Investigate user who has multiple submission errors
We should check mixpanel to investigate [this user](https://stand-with-crypto.sentry.io/issues/4830351975/?environment=production&project=4506490717470720&query=is%3Aunresolved+timesSeen%3A%3E%3D10&referrer=issue-stream&statsPeriod=30d&stream_index=0) and verify that the reason they submitted some many times and failed is because they were entering an invalid address, not because we have a bug in our address flow.

twistershark:
closes #947 

## What changed? Why?

This PR changes the google places api options based on the search results to prevent vague addresses to be used on email flow that would case Capitol Canary to fail when trying to find the related congressperson with the provided address.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/ui/googlePlacesSelect/index.tsx
+++ b/src/components/ui/googlePlacesSelect/index.tsx
@@ -55,15 +55,7 @@ export const GooglePlacesSelect = React.forwardRef<
   // Capitol Canary won't be able to successfully find the related congressperson based on the address provided
   // That's why we need to filter out the route type for US addresses on required action flows
   const usePlacesAutoCompleteResult = shouldLimitUSAddresses
-    ? data
-        ?.map(place =>
-          !place.description.includes('USA')
-            ? place
-            : !place.types.includes('route')
-              ? place
-              : null,
-        )
-        .filter(Boolean)
+    ? data?.filter(place => !place.description.includes('USA') && !place.types.includes('route'))
     : data
 
   const scriptStatus = useGoogleMapsScript()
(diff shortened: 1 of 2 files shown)


src/components/ui/googlePlacesSelect/index.tsx
```diff
@@ -44,6 +50,22 @@ export const GooglePlacesSelect = React.forwardRef<
...
+  const usePlacesAutoCompleteResult = shouldLimitUSAddresses
+    ? data
+        ?.map(place =>
+          !place.description.includes('USA')
+            ? place
+            : !place.types.includes('route')
+              ? place
+              : null,
+        )
+        .filter(Boolean)
```
lucasrmp-cb, line 66: Any reason not to do a single filter here?
```suggestion
    ? data?.filter(place => !place.description.includes('USA') || !place.types.includes('route'))
```

lucasrmp-cb (changes requested):

src/components/ui/googlePlacesSelect/index.tsx
twistershark: Updated! Ty!

olavoparno (approved):

LucasRMP (approved):

lucasrmp-cb (approved):

PR #1004, opened 2024-06-24
Changes caching logic from route to fetch req in all dtsi api route

twistershark:
## What changed? Why?

This PR changes how we are handling cache in all dtsi route handler.

I was investigating on how caching was handle on the api route: /public/dtsi/all-people and I saw that we are defining the route cache as 10 minutes. This would prevent the revalidateTag logic from working because even though the fetch request was being invalidated, the route would keep returning the cache result for 10 minutes.

As we are able to cache the fetch request based on this doc: https://nextjs.org/docs/app/api-reference/functions/fetch, I decided to change the caching for the api route to be completely dynamic and always hitting the fetch function, but as we moved the 10 minutes revalidate prop into the fetch request, it should keep the same "caching" behavior as the route is just going to call the fetch req that is going to be cached and revalidated every 10 minutes.

The idea behind this change is that we'll be able to use revalidateTag to revalidate the fetch request and also keep the revalidation every 10 minutes as well, maintaining the previous behavior and allowing for on-demand fetch revalidation.

## How has it been tested?

I created a new next js app using the app router to test if the fetch cache would work as expected.

Created the following api route to load the data(this is called inside that client home page):
image

this route calls a helper function to fetch the api response. This was not required, but I wanted to make it as similar to how it is happening in SWC as possible:
image

The above function sets the next.tags prop and also sets the revalidate as 10 minutes( 60 sec * 10)

This function calls this api route to get random data, but with the timestamp on when it was requested:
image

I also created a new api route to revalidate the fetch request to test the on-demand revalidation:
image

---

I built the app and opened the same page on two different browsers(chrome and brave), and the response was the same for both of them, even though I accessed the page after a few minutes on brave browser:

image

I'll keep getting the same response for both no matter how many reloads I trigger on each browser until 10 minutes is passed.

image

But if I trigger the revalidateTag, the response is going to be updated for all users on page reload:
image

### DIFF
--- a/src/app/api/public/dtsi/all-people/route.ts
+++ b/src/app/api/public/dtsi/all-people/route.ts
@@ -3,10 +3,9 @@ import 'server-only'
 import { NextResponse } from 'next/server'
 
 import { queryDTSIAllPeople } from '@/data/dtsi/queries/queryDTSIAllPeople'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 
 export const dynamic = 'error'
-export const revalidate = SECONDS_DURATION['10_MINUTES']
+export const revalidate = 0
 
 export async function GET() {
   const data = await queryDTSIAllPeople()
--- a/src/data/dtsi/fetchDTSI.ts
+++ b/src/data/dtsi/fetchDTSI.ts
@@ -57,8 +57,10 @@ export const fetchDTSI = async <R, V = object>(
             query,
             variables,
           }),
-          ...(nextConfig?.nextTags && { next: { tags: nextConfig?.nextTags } }),
-          ...(nextConfig?.nextRevalidate && { next: { revalidate: nextConfig?.nextRevalidate } }),
+          next: {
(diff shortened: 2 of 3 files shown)


src/app/api/public/dtsi/all-people/route.ts
```diff
@@ -3,10 +3,9 @@ import 'server-only'
 import { NextResponse } from 'next/server'
 
 import { queryDTSIAllPeople } from '@/data/dtsi/queries/queryDTSIAllPeople'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 
 export const dynamic = 'error'
-export const revalidate = SECONDS_DURATION['10_MINUTES']
+export const revalidate = 0
```
twistershark, line 8: Changed to 0 based on this doc: https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate

olavoparno (approved):

olavoparno (approved):

lucasrmp-cb (approved):

PR #1021, opened 2024-06-26
feat: goodbye sms journey

LucasdRossi:
closes [#69](https://github.com/Stand-With-Crypto/swc-internal/issues/69)

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/actions/actionUpdateUserProfile.ts
```diff
@@ -141,6 +143,16 @@ async function _actionUpdateUserProfile(data: z.infer<typeof zodUpdateUserProfil
...
+
+    const analytics = getServerAnalytics({
+      localUser,
+      userId: authUser.userId,
+    })
+    await analytics
+      .track('User SMS Opt-In', {
+        provider: 'twilio',
+      })
+      .flush()
```
lucasrmp-cb, line 155: np: if you're doing only one track you can condense this into one statement
```suggestion
    await getServerAnalytics({
      localUser,
      userId: authUser.userId,
    })
      .track('User SMS Opt-In', {
        provider: 'twilio',
      })
      .flush()
```

src/app/api/public/sms/events/messages/route.ts
```diff
@@ -44,7 +54,130 @@ export async function POST(request: NextRequest) {
...
+      phoneNumber,
+    },
+  })
+
+  if (user) {
+    const analytics = getServerAnalytics({
+      localUser: getLocalUserFromUser(user),
+      userId: user.id,
+    })
+    await analytics.track('User SMS Unstop').flush()
```
lucasrmp-cb, line 130: same np here

src/app/api/public/sms/events/messages/route.ts
```diff
@@ -44,7 +54,130 @@ export async function POST(request: NextRequest) {
...
+    const analytics = getServerAnalytics({
+      localUser: getLocalUserFromUser(user),
+      userId: user.id,
+    })
+
+    await analytics
+      .track('User SMS Opt-out', {
+        type: isSWCKeyword ? 'SWC STOP Keyword' : 'STOP Keyword',
+      })
+      .flush()
```
lucasrmp-cb, line 167: and here

src/app/api/public/sms/events/messages/route.ts
```diff
@@ -44,7 +54,130 @@ export async function POST(request: NextRequest) {
...
+    status: 200,
+  })
+}
+
+interface OptUserBackInProps {
+  phoneNumber: string
+  user?: User
+}
+
+async function optUserBackIn({ phoneNumber, user }: OptUserBackInProps) {
```
lucasrmp-cb, line 108: Can we separate this functions into the sms lib? I think this is too much clutter for the route file

src/inngest/functions/capitolCanary/checkSMSOptInReply.ts
```diff
@@ -50,7 +50,11 @@ export const checkSMSOptInReplyWithInngest = inngest.createFunction(
         localUser,
         userId: data.user.id,
       })
-      await analytics.track('User SMS Opt-In').flush()
+      await analytics
+        .track('User SMS Opt-In', {
+          provider: 'capitol-canary',
+        })
+        .flush()
```
lucasrmp-cb, line 57: +1

src/inngest/functions/sms/unstopConfirmationSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  {
+    event: UNSTOP_CONFIRMATION_SMS_COMMUNICATION_JOURNEY_INNGEST_EVENT_NAME,
+  },
+  async ({ event, step }) => {
+    const { phoneNumber } = event.data as UnstopConfirmationSMSCommunicationJourneyPayload
+
+    if (!phoneNumber) {
+      throw new NonRetriableError('Missing phone number')
+    }
```
lucasrmp-cb, line 37: Shouldn't we be checking for a valid phone number here? If the user inputs an invalid phone number I think we should throw this error instead of the 500 that will prob happen when sendgrid tries to send the sms

lucasrmp-cb (changes requested):

src/app/api/public/sms/events/messages/route.ts
```diff
@@ -69,17 +67,15 @@ export async function POST(request: NextRequest) {
       case 'END':
       case 'QUIT':
       case 'STOP':
-        await optOutUser({ phoneNumber, isSWCKeyword: false, user })
-        break
       case SWC_STOP_SMS_KEYWORD:
-        await optOutUser({ phoneNumber, isSWCKeyword: true, user })
+        await optOutUser(phoneNumber, keyword === SWC_STOP_SMS_KEYWORD, user)
```
lucasrmp-cb, line 71: Didn't get this `keyword === SWC_STOP_SMS_KEYWORD` statement, we're inside a case where this is always true, no?

lucasrmp-cb (approved):

src/app/api/public/sms/events/messages/route.ts
LucasdRossi: Not in this scenario, because the case statements above have no break, so line 71 will be executed if the keyword is: "STOPALL" or "UNSUBSCRIBE" or "CANCEL" or ... or SWC_STOP_SMS_KEYWORD

```javascript
const keyword = "STOP"

switch(keyword) {
    case "STOP":
    case "SWC_STOP":
        console.log(keyword === "SWC_STOP") // this will print false
        break
    default:
        break
}
```

lucasrmp-cb (dismissed):

lucasrmp-cb (approved):

PR #1035, opened 2024-06-28
Prepare new call/email flow (for senators)

### LINKED ISSUE #1032: Prepare new call/email flow (for senators)
We need to prepare for a new senate email/call campaign. We can call it SUMMER_2024_SENATE_BILL_CAMPAIGN for now.

We currently [support email/calling house reps, senate reps](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/utils/shared/yourPoliticianCategory.ts#L1), or both in our email and call flows. In anticipation for upcoming activations we should prep a branch to point our email and call flows to senate instead of congress.

Should only need to change a few variables in the [call](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/components/app/userActionFormCallCongressperson/constants.ts#L14) and [email](https://github.com/Stand-With-Crypto/swc-web/blob/main/src/components/app/userActionFormEmailCongressperson/constants.ts#L6) flow and then do some QA to make sure its working as expected

mhamendes:
closes #1032 

## What changed? Why?

- Change call and email flows to target senators

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/cypress/e2e/04-action-email-your-congressperson.cy.ts
+++ b/cypress/e2e/04-action-email-your-congressperson.cy.ts
@@ -3,7 +3,7 @@
 it('action - email your congressperson', () => {
   cy.visit('/')
 
-  cy.contains('Email your congressperson').click()
+  cy.contains(/Email your (congressperson|senator)/).click()
   cy.get('[role="dialog"]')
 
   // validate error messages display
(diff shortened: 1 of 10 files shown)


twistershark:
Add the corresponding title and subtitle for the previous campaings in here: `src/components/app/pageUserProfile/disabledUserActionCampaigns.ts`. 


This is required to show the previous completed campaign in the profile page. If we don't add this when changing the campaign, errors will be sent to Sentry and the profile page will break

olavoparno (dismissed):

olavoparno (dismissed):

lucasrmp-cb (approved):

twistershark (approved):

PR #1047, opened 2024-07-09
Fix: Leaderboard page build performance

### LINKED ISSUE #1044: Improve leaderboard page build performance
For some reason, our build process in Vercel is breaking because the leaderboard static page generation is taking too long that is causing an timeout.

We need to investigate and fix this issue

twistershark:
closes #1044

## What changed? Why?

One of the main reasons why the donation leaderboard page was taking so long to build was related to ENS avatar and name lookups.  As we can't remove the feature of showing the donator ENS information, I added pagination for this page to reduce the amount of ENS lookup calls for each page. We also had an unnecessary DB query that I removed.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/data/aggregations/getSumDonationsByUser.ts
+++ b/src/data/aggregations/getSumDonationsByUser.ts
@@ -1,6 +1,6 @@
 import 'server-only'
 
-import { UserInformationVisibility } from '@prisma/client'
+import { DataCreationMethod, UserInformationVisibility } from '@prisma/client'
 import { Decimal } from '@prisma/client/runtime/library'
 import { compact, keyBy } from 'lodash-es'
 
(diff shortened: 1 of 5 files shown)


olavoparno (dismissed):

src/data/aggregations/getSumDonationsByUser.ts
```diff
@@ -12,16 +12,47 @@ import { prismaClient } from '@/utils/server/prismaClient'
 export type SumDonationsByUserConfig = {
   limit: number
   offset?: number
+  pageNum: number
 }
 
-async function getSumDonationsByUserQuery({ limit, offset }: SumDonationsByUserConfig) {
-  const total: {
+type GetSumDonationByUserQueryResult = Array<{
```
travisbloom-cb, line 18: I would have the prisma req as its own fn and then reference the return type of that fn, rather than manually keeping it in sync

src/data/aggregations/getSumDonationsByUser.ts
```diff
@@ -35,33 +66,22 @@ async function getSumDonationsByUserQuery({ limit, offset }: SumDonationsByUserC
...
     totalAmountUsd: totalDonationAmountUsd.toNumber(),
+    ...rest,
   }))
 }
 
 type QueryResult = Awaited<ReturnType<typeof getSumDonationsByUserQuery>>
 
 // If we ever have an nft mint action that is not a "donation", we'll need to refactor this logic
 async function getSumDonationsByUserData(total: QueryResult) {
-  const users = await prismaClient.user.findMany({
```
travisbloom-cb, line 48: I have to image this wasn't a big perf issue?

twistershark: Yeah... this db query was taking a few hundred ms because it used index search instead of a table scan, so the impact was not so significant, but we already had a query that requested all that information, so it wasn't really necessary to request the data in another query again.

src/data/aggregations/getSumDonationsByUser.ts
twistershark: We actually already have this on line https://github.com/Stand-With-Crypto/swc-web/blob/b23ba8c9e0aa73b72bbe4831bfc9b2dd1dde3170/src/data/aggregations/getSumDonationsByUser.ts#L51, so we don't need to explicit define the type as the total constant is getting the type from prisma.

travisbloom-cb (approved):

src/data/aggregations/getSumDonationsByUser.ts
twistershark: image

olavoparno (approved):

PR #1052, opened 2024-07-10
Fix: Leaderboard page build timeout

### LINKED ISSUE #1044: Improve leaderboard page build performance

twistershark:
closes #1044 

## What changed? Why?

This PR creates a new approach for returning default values for promises inside an Promise.all when a defined maximum timeout is reached. The idea here is to prevent the build page worker on Vercel from killing the build process because a promise was not resolved during the default 60 seconds limit on the Leaderboard page.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/data/web3/getENSDataFromCryptoAddress.ts
+++ b/src/data/web3/getENSDataFromCryptoAddress.ts
@@ -4,6 +4,8 @@ import { compact } from 'lodash-es'
 import { formatENSAvatar } from '@/utils/server/formatENSAvatar'
 import { thirdwebRPCClient } from '@/utils/server/thirdweb/thirdwebRPCClients'
 import { IS_DEVELOPING_OFFLINE } from '@/utils/shared/executionEnvironment'
+import { raceAll } from '@/utils/shared/raceAll'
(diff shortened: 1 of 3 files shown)


src/data/web3/getENSDataFromCryptoAddress.ts
```diff
@@ -14,27 +16,33 @@ async function _getENSDataMapFromCryptoAddresses(
...
     .filter(({ ensName }) => ensName)
-  const records = await Promise.all(
+
+  const records = await raceAll(
     addressesWithENS.map(address =>
       client.getEnsAvatar({
         name: address.ensName!,
       }),
     ),
+    SECONDS_DURATION['30_SECONDS'] * 1000,
```
travisbloom-cb, line 43: isnt this way more than 30 seconds?

travisbloom-cb (approved):

PR #1055, opened 2024-07-10
feat: bulk sms journey

LucasdRossi:
closes [#64](https://github.com/Stand-With-Crypto/swc-internal/issues/64)

fixes 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,150 @@
...
+    const { smsBody, userWhereInput } = event.data as BulkSMSCommunicationJourneyPayload
+
+    if (!smsBody) {
+      throw new NonRetriableError('Missing sms body')
+    }
+
+    const segmentsCount = countSegments(smsBody)
+
+    const userBatches = await step.run('fetch-users', async () => {
+      const phoneNumbers = await prismaClient.user.groupBy({
```
travisbloom-cb, line 47: prisma limits to 100k rows in queries so you might need to paginate your queries

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      return batches
+    })
+
+    let batchNum = 1
+    let messageCount = 0
+    for (const batch of userBatches) {
+      await Promise.all(
+        batch.map((phoneNumber, index) =>
+          step.run(`send-sms-${batchNum}:${index + 1}`, async () => {
```
travisbloom-cb, line 107: this could get expensive at scale (https://www.inngest.com/pricing?ref=nav) if we're sending texts to 500k+ users at once. We might want to rethink where we use steps

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,150 @@
...
+            if (message) {
+              messageCount += 1
+              await createCommunication(communicationJourneys, message.sid)
+            }
+          }),
+        ),
+      )
+
+      batchNum += 1
+      await step.sleep(`script.sleep-${batchNum}`, 1000 * segmentsCount)
```
travisbloom-cb, line 127: Why?

src/inngest/functions/sms/shared/validatePhoneNumber.ts
```diff
@@ -1,13 +1,33 @@
...
   if (!PHONE_NUMBER_REGEX.test(phoneNumber)) {
     throw new NonRetriableError('Invalid phone number')
   }
+
+  // only send SMS to US and CA numbers
+  if (!phoneNumber.startsWith('+1')) {
+    throw new NonRetriableError('Phone number not from US or CA')
+  }
+
+  const lookup = await messagingClient.lookups.v2.phoneNumbers(phoneNumber).fetch()
```
travisbloom-cb, line 20: This is super pricey https://www.twilio.com/en-us/verify/pricing. We prob want to track if texting a number fails (because its an invalid number), and then flag in the db that the number shouldnt be texted. Rather than always validating using this API

travisbloom-cb (changes requested):

src/inngest/functions/sms/shared/validatePhoneNumber.ts
LucasdRossi: The lookup API is free https://www.twilio.com/en-us/user-authentication-identity/pricing/lookup. We'll only get charged if we use additional features like: Inbound caller identification, Identity match and Reassigned number risk.

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
LucasdRossi: With an MPS of 3 Twilio will send 3 segments each second, so if the message has more than one segment we need to wait for all segments to be sent before putting another SMS in queue. We need to do this because we're using a toll-free number and it has max queue of 108,000 so if we don't take the number of segments in consideration and keep adding SMS to the queue, the queue will get full and messages won't be send.

I'll add a comment explaining this

src/inngest/functions/sms/shared/validatePhoneNumber.ts
LucasdRossi: I agree that we should flag invalid numbers in the db, but how should we do it?
1. Set smsStatus to NOT_OPTED_IN; or
2. Add a new enum to the smsStatus field; or
3. Add a new flag to the user's table; or
4. Clear the phoneNumber input?

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
LucasdRossi: Ok, maybe we should fill the queue, wait for it to be empty and repeat this until there are no more messages. This way we would reduce the number of steps. What do you think?

src/inngest/functions/sms/shared/validatePhoneNumber.ts
LucasdRossi: Created a new issue to work on this [#90](https://github.com/Stand-With-Crypto/swc-internal/issues/90)

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,356 @@
...
+      'time-estimation',
+      async () => {
+        logger.info('Estimating...')
+
+        let totalLength = 0
+        let cursor: Date | undefined
+        let hasNumbersLeft = true
+
+        // We could do this using two other methods:
+        // count() with distinct: not supported https://github.com/prisma/prisma/issues/4228. Would need a workaround that doesn't support custom userWhereInput
```
travisbloom-cb, line 81: we could just write a `prisma.rawQuery` for count distinct? Fine either way if this works

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,356 @@
+import { Prisma, SMSStatus, User, UserCommunicationJourneyType } from '@prisma/client'
```
travisbloom-cb, line 1: I wonder if we could break some of the computation logic in to unit-testable functions

travisbloom-cb (dismissed):

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
LucasdRossi: Yes we could, but then we wouldn't be able to take in count the userWhereInput, because it's a prisma object.

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,356 @@
...
+    return `${minutes} minutes`
+  } else if (seconds < 86400) {
+    // less than 24 hours (86400 seconds)
+    const hours = Math.ceil(seconds / 3600)
+    return `${hours} hours`
+  } else {
+    // 24 hours or more
+    const days = Math.ceil(seconds / 86400)
+    return `${days} days`
+  }
```
lucasrmp-cb, line 324: We have a seconds duration shared constant for this kind of comparison
```suggestion
  if (seconds < SECONDS_DURATION.MINUTE) {
    return `${seconds.toPrecision(2)} seconds`
  } else if (seconds < SECONDS_DURATION.HOUR) {
    // less than 60 minutes (3600 seconds)
    const minutes = Math.ceil(seconds / SECONDS_DURATION.MINUTE)
    return `${minutes} minutes`
  } else if (seconds < SECONDS_DURATION.DAY) {
    // less than 24 hours (86400 seconds)
    const hours = Math.ceil(seconds / SECONDS_DURATION.HOUR)
    return `${hours} hours`
  } else {
    // 24 hours or more
    const days = Math.ceil(seconds / SECONDS_DURATION.DAY)
    return `${days} days`
  }
```

src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -0,0 +1,354 @@
...
+)
+const MAX_QUEUE_LENGTH = Number(requiredEnv(process.env.MAX_QUEUE_LENGTH, 'MAX_QUEUE_LENGTH'))
+
+// This constant is specific to our twilio account
+const QUEUING_THROUGHPUT = 500
+
+const logger = getLogger('bulk-sms')
+
+interface BulkSMSCommunicationJourneyPayload {
+  smsBody: string
```
travisbloom-cb, line 39: we'll want a persist flag as well so we can run this and log out how many people are gonna get texted without actually triggering (to audit before sending)

lucasrmp-cb (approved):

travisbloom-cb (approved):

PR #1061, opened 2024-07-12
feat: add captureException to rate limit error throw

### LINKED ISSUE #1060: Improve rate limit error logging

lucasrmp-cb:
closes #1060 

fixes 

## What changed? Why?

This is just adding a sentry log so we have more info on rate limit errors

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/server/ratelimit/throwIfRateLimited.ts
+++ b/src/utils/server/ratelimit/throwIfRateLimited.ts
@@ -1,3 +1,5 @@
+import * as Sentry from '@sentry/nextjs'
+
 import { getIPFromHeaders } from '@/utils/server/getIPFromHeaders'
 import {
   authenticatedRateLimiter,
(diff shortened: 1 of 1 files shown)


src/utils/server/ratelimit/throwIfRateLimited.ts
```diff
@@ -21,6 +23,18 @@ export async function throwIfRateLimited({
 
   const result = await ratelimiter.limit(ip)
   if (!result.success) {
+    Sentry.captureException('Rate limit exceeded', {
+      user: {
+        ip_address: ip,
```
mhamendes, line 28: Is the IP being captured? Sentry has an option to prevent logging IP addresses and it can be a GPDR concern, no?

lucasrmp-cb: We're using the IP as user identification, so it shouldn't block it.
Maybe it will be redacted from the "Extra information" section, but not from the user identification

travisbloom-cb (approved):

PR #1073, opened 2024-07-16
Disable all actions from working for non-US users

eduardopicolo-cb:
## What changed? Why?

Block non-US users from triggering actions/signing up & show a disclaimer banner letting them know actions are disabled.

image
image

## Notes to reviewers

There's an option to change your user's country code `/internal/user-settings`. VPN also works.

For local testing, there is a new env var `USER_COUNTRY_CODE`. This var is used only on `local/testing` environments and if it's missing, it will default to `'US'`.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [x] Unit test
- [ ] Functional test


src/components/app/geoGate.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const GeoGate = (props: GeoGateProps) => {
+  const { children, countryCode, unavailableContent, bypassCountryCheck } = props
+
+  const hasHydrated = useHasHydrated()
+
+  const userCountryCode = Cookies.get(USER_COUNTRY_CODE_COOKIE_NAME)
+
+  if (!hasHydrated) return children
```
travisbloom-cb, line 22: 🙌

src/components/app/userActionFormCommon/actionUnavailable.tsx
```diff
@@ -0,0 +1,25 @@
+import { Button } from '@/components/ui/button'
+import { InternalLink } from '@/components/ui/link'
+import { PageSubTitle } from '@/components/ui/pageSubTitle'
+import { PageTitle } from '@/components/ui/pageTitleText'
+
+interface UserActionFormActionUnavailableProps {
+  onConfirm?: () => void
+}
+
+export const UserActionFormActionUnavailable = (props: UserActionFormActionUnavailableProps) => {
```
travisbloom-cb, line 10: Lets trigger an analytics event every time this is displayed

src/components/app/userActionFormCommon/dialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+    children,
+    trigger,
+    countryCode = DEFAULT_SUPPORTED_COUNTRY_CODE,
+    bypassCountryCheck = false,
+    padding = true,
+    ...dialogProps
+  } = props
+
+  const unavailableContent = (
+    <Dialog {...dialogProps} analytics={ANALYTICS_NAME_USER_ACTION_FORM_UNAVAILABLE}>
```
travisbloom-cb, line 30: Is there a reason we're rendering different dialogs? could we reduce jank if we have the Geogate nested in the parent dialog?

src/utils/server/setCookie.ts
```diff
@@ -0,0 +1,7 @@
+'use server'
+
+import { cookies } from 'next/headers'
+
+export async function setCookie(...args: Parameters<ReturnType<typeof cookies>['set']>) {
```
travisbloom-cb, line 5: are we using?

src/utils/server/userActionValidation/checkCountryCode.ts
```diff
@@ -0,0 +1,28 @@
...
+import { USER_COUNTRY_CODE_COOKIE_NAME } from '@/utils/server/getCountryCode'
+import { UserActionValidationErrors } from '@/utils/server/userActionValidation/constants'
+
+export const checkCountryCode = async (countryCode: string) => {
+  const userCountryCode = cookies().get(USER_COUNTRY_CODE_COOKIE_NAME)?.value
+  return userCountryCode === countryCode
+}
+
+export function createCountryCodeValidation(requiredCountryCode: string) {
+  return async () => {
```
travisbloom-cb, line 14: any reason this is async?

travisbloom-cb (commented):
This looks great!

src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
```diff
@@ -163,18 +165,25 @@ export function SuggestedScript({
...
             'DTSI Slug': data.dtsiSlug,
           },
           payload: data,
+          onError: (key, error) => {
+            if (key === UserActionValidationErrors.ACTION_UNAVAILABLE) {
+              toast.error('Action unavailable', {
+                description: error.message,
+              })
+            }
+            toastGenericError()
```
lucasrmp-cb, line 181: If this `if` condition is met you're showing 2 toasts, don't think this is expected

src/components/app/userActionFormNFTMint/sections/transactionWatch.tsx
```diff
@@ -63,7 +65,14 @@ export function UserActionFormNFTMintTransactionWatch({
...
       {
         formName: 'User Action Form NFT Mint',
-        onError: toastGenericError,
+        onError: (key, error) => {
+          if (key === UserActionValidationErrors.ACTION_UNAVAILABLE) {
+            toast.error('Action unavailable', {
+              description: error.message,
+            })
+          }
+          toastGenericError()
```
lucasrmp-cb, line 74: Same problem with duplicated toasts here

src/utils/server/setCookie.ts
lucasrmp-cb: Unnecessary abstraction IMO

lucasrmp-cb (commented):
Looking good

src/utils/server/setCookie.ts
eduardopicolo-cb: Yes, for the /internal/user-config page

src/utils/server/userActionValidation/checkCountryCode.ts
eduardopicolo-cb: Its a server-action, so it has to be async otherwise nextjs throws an error

src/utils/server/setCookie.ts
eduardopicolo-cb: Due to be `cookies` being a dynamic function, it cannot be directly imported in a client component:
`Error: 
  × You're importing a component that needs next/headers. That only works in a Server Component but one of its parents is marked with "use client"`
  
This is only being used internally though.

src/components/app/pageDonate/donateButton.tsx
```diff
@@ -15,12 +16,18 @@ export function DonateButton() {
     await triggerServerActionForForm(
       {
         formName: 'Donate Button',
-        onError: toastGenericError,
+        onError(key, error) {
+          if (key === UserActionValidationErrors.ACTION_UNAVAILABLE) {
+            toast.error('Action unavailable', {
+              description: error.message,
+            })
+          }
```
lucasrmp-cb, line 24: We still want to toast a generic error when this doesn't match, but it should be an `else` statement or you should have a return in this block

src/components/app/userActionFormCallCongressperson/sections/suggestedScript.tsx
```diff
@@ -163,18 +165,24 @@ export function SuggestedScript({
...
             'User Action Type': UserActionType.CALL,
             'DTSI Slug': data.dtsiSlug,
           },
           payload: data,
+          onError: (key, error) => {
+            if (key === UserActionValidationErrors.ACTION_UNAVAILABLE) {
+              toast.error('Action unavailable', {
+                description: error.message,
+              })
+            } else toastGenericError()
```
lucasrmp-cb, line 180: Like this

lucasrmp-cb (changes requested):
Comments above

src/utils/server/userActionValidation/withValidations.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function withValidations<T extends (...args: any) => any>(
+  validations: ValidationFunction<T>[],
+  action: T,
+) {
+  return async (
+    ...args: Parameters<T>
+  ): Promise<Awaited<ReturnType<T> | ReturnType<ValidationFunction<T>>>> => {
+    for (const validation of validations) {
+      const result = await validation(args)
```
travisbloom-cb, line 13: we should probably:
- throw errors here
- log to sentry

travisbloom-cb (commented):
@eduardopicolo-cb doesn't this mean we won't show a descriptive error anymore?

eduardopicolo-cb:
> @eduardopicolo-cb doesn't this mean we won't show a descriptive error anymore?

We won't. But a geo gated client should never be able to trigger an action

travisbloom-cb (approved):

Issue #1053, opened 2024-07-10
Replace recent activity with advocates map at homepage


olavoparno:
replace recent activity data with the map from https://www.standwithcrypto.org/map while adjusting some styling

https://www.figma.com/design/MWzmEAYQC1NYZiD4BG0FlI/V2-Fast-Follows?node-id=4399-2735&t=RWGFix0GV7GqPCCE-0

key notes: 
- remove our mission section
- on mobile keep activity data list instead of map
- on desktop show map instead of activity data list

olavoparno:
- [ ] Reduce spacing between the header and the body text?
- [ ] Update all of the section headers (our community, our partners, get involved, etc) to be 32px instead of 48px?
- [ ] Have that last line “Donations to Fairshake, a pro-crypto Super PAC, are not included on the leaderboard.” only show up if the user is on the “Top donations” tab and not the “Recent activity” tab?
- [ ] [Change logic on map to pull last 20 actions that have a location](https://github.com/Stand-With-Crypto/swc-internal/issues/93)

PR #1078, opened 2024-07-18
Feat/add advocate map at home

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1053

## How has it been tested?

- [X] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/data/recentActivity/getPublicRecentActivity.ts
```diff
@@ -20,19 +20,11 @@ const fetchFromPrisma = async (config: RecentActivityConfig) => {
     orderBy: {
       datetimeCreated: 'desc',
     },
-    take: config.limit,
+    take: 1000,
```
travisbloom-cb, line 23: we only want this if `config.restrictToUS` is passed

src/data/recentActivity/getPublicRecentActivity.ts
```diff
@@ -55,9 +47,20 @@ const fetchFromPrisma = async (config: RecentActivityConfig) => {
 }
 
 export const getPublicRecentActivity = async (config: RecentActivityConfig) => {
-  const data = await fetchFromPrisma(config)
+  const rawData = await fetchFromPrisma(config)
   const dtsiSlugs = new Set<string>()
```
travisbloom-cb, line 52: lets add comments explaining why we're doing this in memory and add a TODO to figure out how to do it performantly via the db

travisbloom-cb (changes requested):

travisbloom-cb (approved):

PR #1079, opened 2024-07-18
Migrate Thirdweb to V5

### LINKED ISSUE #773: Migrate to v5 of Thirdweb SDK

twistershark:
closes #773 

## What changed? Why?

This PR migrates the codebase to work with thirdweb on V5.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/userActionFormCommon/nftDisplay.tsx
```diff
@@ -41,23 +44,33 @@ export function NFTDisplay({
...
-          height={NFT_IMAGE_SIZE_BY_VARIANT[size]}
-          width={NFT_IMAGE_SIZE_BY_VARIANT[size]}
-        />
-      )}
-    </div>
+  if (loading) {
+    return <LoadingOverlay size="sm" />
+  }
+
+  const image = isThirdwebMedia ? (
```
olavoparno, line 54: i'd just try avoiding nested ternaries here man, maybe abstracting it into a function getImage with early returns

src/components/app/authentication/loginDialogWrapper.tsx
```diff
@@ -203,6 +210,11 @@ export function UnauthenticatedSection({
     <Dialog {...dialogProps} onOpenChange={setDialogOpen}>
       <DialogTrigger asChild>{children}</DialogTrigger>
       <DialogContent className="max-w-l w-full">
+        <VisuallyHidden>
+          <DialogTitle>
```
olavoparno, line 214: does it even make sense having DialogTitle or any other non primitive tags inside VisuallyHidden? just wondering if I got it right

olavoparno (approved):

olavoparno:
nice work, man

src/components/app/userActionFormCommon/nftDisplay.tsx
twistershark: Fixed!

src/components/app/authentication/loginDialogWrapper.tsx
twistershark: The idea here is to remove the error message related to accessibility and missing the DialogTitle inside DialogContent.

To prevent showing the DialogTitle, I'm using this VisuallyHidden component from RadixUI to hide it for normal users, but allow the content to be read by screen readers

olavoparno (dismissed):

src/components/app/userActionFormNFTMint/useCheckoutController.ts
```diff
@@ -26,6 +29,7 @@ export function useCheckoutController({
 } = {}): UseCheckoutControllerReturn {
   const [quantity, setQuantity] = React.useState(1)
   const { data: gasFee } = useGasFee(quantity)
+  const gasNumberInBigNumber = BigNumber.from(0).add(gasFee ?? 0)
```
lucasrmp-cb, line 32: wouldn't `BigNumber.from(gasFee)` work?

src/components/app/userActionFormNFTMint/useCheckoutController.ts
```diff
@@ -58,14 +62,20 @@ export function useCheckoutController({
...
+
+  return useSWR({ contract }, async ({ contract: _contract }) => {
     if (!contract) {
       return
     }
-    const prepareTx = await contract.erc721.claim.prepare(quantity)
-    prepareTx.updateOverrides({ gasLimit: CLAIM_GAS_LIMIT_OE721_CONTRACT })
-    const gasFee = await prepareTx.estimateGasCost()
+
+    const tx = claimTo({ contract, quantity: BigInt(quantity), to: account?.address ?? '' })
```
lucasrmp-cb, line 77: Is this actually creating the transaction or is it just preparing it?

src/hooks/useAuthUser.ts
```diff
@@ -1,25 +1,21 @@
...
-import { AuthSessionMetadata } from '@/utils/server/thirdweb/types'
+import { THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX } from '@/utils/shared/thirdwebAuthToken'
 
 export function useThirdwebAuthUser() {
-  const data = useUser<Json, AuthSessionMetadata>()
+  const token = Cookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)
+  const decodedToken = token
+    ? jwtDecode<{ ctx?: { userId?: string; address?: string } }>(token)
+    : null
+  const { userId, address } = decodedToken?.ctx ?? {}
```
lucasrmp-cb, line 12: This should be inside a `useMemo`, we're doing a cookie get and a jwt decode, we don't want to do it on every render

src/hooks/useENS.ts
```diff
@@ -0,0 +1,35 @@
...
+    ens,
+    avatarUrl,
+  }
+}
+
+export function useENS() {
+  const { user } = useThirdwebAuthUser()
+  const address = user?.address
+
+  return useSWR(address, async () => await getData(address))
```
lucasrmp-cb, line 34: Not the optimal way to do this, we don't need to access `address` from the scope inside the fetcher function
```suggestion
  return useSWR(user?.address, getData)
```
What you pass to the first arg of `useSWR` is passed as arg to the fetcher fn, so just passing getData would access the value in the way `swr` intends

src/hooks/useSendMintNFTTransaction.ts
```diff
@@ -1,145 +1,63 @@
...
-  onStatusChange = noop,
 }: UseSendMintNFTTransactionOptions) {
-  const sdk = useSDK()
-  const { contract } = useContract(contractAddress)
-  const { data: contractMetadata } = useContractMetadata(contract)
+  const contract = getContract({
+    address: contractAddress,
+    client: thirdwebClient,
+    chain: base,
+  })
```
lucasrmp-cb, line 28: Don't know what this does, but smells to me like it should be in a `useMemo`

lucasrmp-cb (changes requested):
Looking good

src/components/app/userActionFormNFTMint/useCheckoutController.ts
twistershark: Actually it doesn't. We can't just use `BigNumber.from(gasFee)` because they expect a number or string for the from method and gasFee is a bigint. Strangely, add accepts bigint. So the idea here is to create a new BigNumber instance as zero and then adding the gasFee(bigint) value

src/hooks/useSendMintNFTTransaction.ts
twistershark: This function is a replacement for the old useContract hook, allowing us to get contract information without needing to be inside a component or hook

twistershark: Added the useMemo

src/hooks/useENS.ts
twistershark: Thanks for the suggestion! Updated!

src/hooks/useAuthUser.ts
twistershark: Updated

src/components/app/userActionFormNFTMint/useCheckoutController.ts
twistershark: Just preparing the transaction to get the current gasFee price, but we don't use this prepared transaction afterwards as we recreate it again on the `onMint` function that actually prepares the transaction to be executed

src/hooks/useAuthUser.ts
```diff
@@ -1,25 +1,30 @@
...
+  const token = Cookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)
+
+  const { userId, address } = useMemo(() => {
+    const decodedToken = token
+      ? jwtDecode<{ ctx?: { userId?: string; address?: string } }>(token)
+      : null
+    const { userId, address } = decodedToken?.ctx ?? {}
+
+    return { userId, address }
+  }, [token])
```
lucasrmp-cb, line 21: I think we merged two possible solution we discussed here, not sure that both are necessary, try one of these:
```suggestion
  const loginStatus = useActiveWalletConnectionStatus()

  const { userId, address } = useMemo(() => {
    const token = Cookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)
    const decodedToken = token
      ? jwtDecode<{ ctx?: { userId?: string; address?: string } }>(token)
      : null
    const { userId, address } = decodedToken?.ctx ?? {}

    return { userId, address }
  }, [loginStatus])
```

**OR**

```suggestion
  const token = Cookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)

  const { userId, address } = useMemo(() => {
    const decodedToken = token
      ? jwtDecode<{ ctx?: { userId?: string; address?: string } }>(token)
      : null
    const { userId, address } = decodedToken?.ctx ?? {}

    return { userId, address }
  }, [token])
```

lucasrmp-cb: Both of them simplify this code in a different way. If neither work please add a comment to the `useActiveWalletConnectionStatus` call explaining why we aren't using the return value

twistershark: I can't use the `loginStatus` as a dependency on useMemo because React detects that it is not necessary and causes error on the lint verification step. I added a comment with why we are using this `useActiveWalletConnectionStatus` hook without the returned values.

olavoparno (dismissed):

lucasrmp-cb (approved):

olavoparno (dismissed):

olavoparno (approved):

PR #1162, opened 2024-08-07
Events page BuilderIO integration

twistershark:
## What changed? Why?

This PR integrates the events page with builder.io

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/app/[locale]/events/[state]/[eventSlug]/page.tsx
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/[eventSlug]/page.tsx:1]: Do we really need the state for this? Since the slug is pseudo-unique, can't we make this route `/events/[eventSlug]`?

src/app/[locale]/events/[state]/[eventSlug]/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const event = await getEvent(eventSlug, state)
+
+  const isStateValid = Object.keys(US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP).includes(
+    state.toUpperCase(),
+  )
+
+  if (!isStateValid || !event) {
+    permanentRedirect(intlUrls.events())
+  }
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/[eventSlug]/page.tsx:40]: I think we want a sentry log or a 404 page here (I lean towards a sentry log)

src/app/[locale]/events/[state]/[eventSlug]/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+  'Stand With Crypto hosts events nationwide to organize, activate, and energize the nationwide Crypto community. Check this page for information about upcoming events, including times, dates, and locations, and RSVP to events in your area'
+
+export const metadata: Metadata = {
+  ...generateMetadataDetails({
+    title,
+    description,
+  }),
+}
+
+export default async function EventDetailsPageRoot({ params }: Props) {
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/[eventSlug]/page.tsx:28]: Can this page be static with a revalidate?

src/app/[locale]/events/[state]/page.tsx
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/page.tsx:1]: Realizing that `/events/[eventSlug]` would clash with this, but I'd rather change this to `/events/by-state/[state]` (or something like that), that way the vanity url for the event details page would be cleaner

src/app/[locale]/events/[state]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { state, locale } = params
+  const intlUrls = getIntlUrls(locale)
+
+  const isStateValid = Object.keys(US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP).includes(
+    state.toUpperCase(),
+  )
+
+  if (!isStateValid) {
+    permanentRedirect(intlUrls.events())
+  }
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/page.tsx:35]: Let's add a sentry log here

src/app/[locale]/events/[state]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+export async function generateMetadata({ params }: Props): Promise<Metadata> {
+  const { state } = params
+
+  return generateMetadataDetails({
+    title: `Events in ${US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP[state.toUpperCase() as keyof typeof US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP]}`,
+    description,
+  })
+}
+
+export default async function StateEventsPageRoot({ params }: Props) {
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/[state]/page.tsx:25]: Same point on making this static with a revalidate, we usually want to reduce CMS calls as much as possible, having these pages with `revalidate` makes so we don't do one builder io request per user access

src/app/[locale]/events/page.tsx
```diff
@@ -16,6 +16,12 @@ export const metadata: Metadata = {
...
 }
 
-export default function CNNDebatePage() {
-  return <EventsPage />
+export default async function EventsPageRoot() {
+  const events = await getEvents()
+
+  if (!events || !events?.length) {
+    return NotFound()
+  }
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/page.tsx:24]: lets add a sentry log here too

src/app/[locale]/events/page.tsx
```diff
@@ -16,6 +16,12 @@ export const metadata: Metadata = {
   }),
 }
 
-export default function CNNDebatePage() {
-  return <EventsPage />
+export default async function EventsPageRoot() {
+  const events = await getEvents()
+
+  if (!events || !events?.length) {
+    return NotFound()
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/page.tsx:23]: We should use the `notFound()` function from `react/navigation` here, not this one

src/app/[locale]/events/page.tsx
```diff
@@ -16,6 +16,12 @@ export const metadata: Metadata = {
...
+  }
+
+  return <EventsPage events={events!} />
```
Comment (lucasrmp-cb) [inline src/app/[locale]/events/page.tsx:26]: np: is this `!` necessary?

lucasrmp-cb (changes requested):

olavoparno (approved):

src/app/[locale]/events/[state]/[eventSlug]/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/[state]/[eventSlug]/page.tsx:1]: Actually it won't be unique because of that builder up limitation. I'm using the state to try to be more specific

src/app/[locale]/events/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/page.tsx:26]: Yes... even though I checked above if it was not null, TS was still complaining, but at this point we are sure that it is not null

src/app/[locale]/events/[state]/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/[state]/page.tsx:1]: Decided to used the state to try to make is a bit more specific, trying to overcome the builder up slug limitation. This way, if we had the same even happening on two states, if the admin reused the same slug, at least the state in the router would help making it load the correct event. But I can change this no problem. I'm fine with whatever

src/app/[locale]/events/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/page.tsx:23]: Updated!

src/app/[locale]/events/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/page.tsx:24]: Do we need to manually add it here? Because our not-found page already sends this event to Sentry

src/app/[locale]/events/[state]/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/[state]/page.tsx:25]: Updated

src/app/[locale]/events/[state]/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/[state]/page.tsx:35]: Instead or redirecting, I decided to render the notFound which already has the sentry log

src/app/[locale]/events/[state]/[eventSlug]/page.tsx
Comment (twistershark) [reply src/app/[locale]/events/[state]/[eventSlug]/page.tsx:28]: Updated

lucasrmp-cb (approved):

Issue #1109, opened 2024-07-26
Support reactivation transactional emails


travisbloom-cb:
PRD: https://docs.google.com/document/d/1xJDpcJDKAlEtEw2KReUZs4DJutk3hiOJuXtFOk-nlu0/edit 
Figma: https://www.figma.com/design/NYBTFMwD8iL9ZYwpfmqhRv/Transactional-Emails?node-id=1-17&m=dev

This task is meant to send welcome emails to those who haven't received it yet.

travisbloom-cb:
@lucasrmp-cb to provide onboarding to @olavoparno for this feature

PR #1171, opened 2024-08-09
Feat/add reactivation email and backfill

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1109

## What changed? Why?

- inngest function to send welcome email to users who havent received it yet

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/inngest/functions/backfillReactivation/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  { event: BACKFILL_REACTIVATION_INNGEST_EVENT_NAME },
+  async ({ event, step }) => {
+    const payload = event.data as ScriptPayload
+    return await step.run('execute-script', async () => {
+      return await backfillReactivation({
+        limit: payload.limit,
+        persist: payload.persist,
+      })
+    })
+  },
```
mhamendes, line 28: Did you confirm that this backfill completes before the function times out? If I am not mistaken, Vercel serverless functions times out after 90s.

olavoparno: I haven't thought of that to be honest. The worst case scenario I think is having to run it again, but I'll take a deeper look at it. Thanks man

src/utils/server/email/templates/reactivationReminder.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </Text>
+        <br />
+        <br />
+        <Text className="text-foreground-muted my-0 text-center text-[15px]">
+          Stay tuned for more updates and information from SWC. You can always visit{' '}
+          <a
+            className="text-foreground-muted"
+            href="https://www.standwithcrypto.org"
+            target="_blank"
+          >
```
lucasrmp-cb, line 62: 2 things:
- We should always use react-email's built in components whenever possible, and they have a anchor component called `Link`
- We should always use `buildTemplateInternalUrl` for links inside stand with crypto, so this would end up like this
```suggestion
          <Link className="text-inherit underline" href={buildTemplateInternalUrl('/', hydratedHrefSearchParams)}>
            our website
          </Link>
```

lucasrmp-cb (changes requested):

src/utils/server/email/templates/reactivationReminder.tsx
olavoparno: thank you man, I've adjusted!

lucasrmp-cb (approved):

PR #1174, opened 2024-08-09
Completes Events Page

twistershark:
## What changed? Why?

This PR completes the events page by creating the notification logic, sharing links, events map and other improvement.s

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/pageEvents/components/eventDialogContent.tsx
```diff
@@ -1,56 +1,152 @@
...
+    })
+
+    toast.success('Thank you for your interest! We will send you a reminder closer to the event.')
+    sectionProps.goToSection(SectionNames.NOTIFICATION_ACTIVATED)
+  }
+
+  async function handleRSVPButtonClick() {
+    await handleCreateRsvpAction({
+      shouldReceiveNotifications: false,
+    })
```
mhamendes, line 110: Here is what we talked about making it async. Just change to:

```suggestion
    void handleCreateRsvpAction({
      shouldReceiveNotifications: false,
    })
```

This will trigger the create action but will not block the code.

src/components/app/pageEvents/components/eventsMap.tsx
```diff
@@ -0,0 +1,245 @@
+'use client'
```
mhamendes, line 1: We should probably create a generic component for the map in the future so we don't have to copy paste the entire map logic when we need to add it to another component and with other data.

olavoparno (dismissed):

src/components/app/pageEvents/components/stateEventsDialogContent.tsx
```diff
@@ -39,30 +43,33 @@ export function StateEventsDialogContent({ state, events }: StateEventsDialogPro
...
-            <EventDialog
-              event={event.data}
-              key={event.data.slug}
-              trigger={<StateDialogEventCard event={event.data} />}
-            />
-          ))}
-        </div>
-      )}
+      <ScrollArea>
+        {stateEvents && stateEvents?.length > 0 && (
```
mhamendes, line 47: This should be enough:

```suggestion
        {stateEvents?.length && (
```

src/inngest/functions/eventNotification/logic.ts
```diff
@@ -0,0 +1,115 @@
...
+        shouldReceiveNotifications: true,
+      },
+      include: {
+        userAction: {
+          include: {
+            user: true,
+          },
+        },
+      },
+    })
```
mhamendes, line 81: Prisma limits the amount of rows to 100k. If that is a case where we have more than 100k events to notify, this will break. In other cases we create batches to do it.

src/inngest/functions/eventNotification/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  {
+    id: SEND_EVENT_NOTIFICATION_INNGEST_FUNCTION_ID,
+    retries: 0,
+    onFailure: onScriptFailure,
+  },
+  { cron: 'TZ=America/New_York 0 12 * * *' }, // Every day - 12PM EST
+  async ({ step }) => {
+    return await step.run('execute-script', async () => {
+      return await sendEventNotifications()
+    })
```
mhamendes, line 16: Its only one step on this run. Vercel has a limit of 3 minutes for a serverless function to resolve, if it takes longer than that, it will break.

src/components/app/pageEvents/components/eventsMap.tsx
twistershark: This is actually something I talked about with Olavo. We will do this refactor in the future, but doing it in this PR would create too much information at the same PR. Great suggestion!

src/inngest/functions/eventNotification/logic.ts
twistershark: Totally agree. This is something I talked with Bernardo about and I think that for now we should be fine as I don't think that we will have more than 100k notifications to send per event

src/components/app/pageEvents/components/stateEventsDialogContent.tsx
lucasrmp-cb: image
Don't do that, `stateEvents?.length > 0` is the most compact and safe way to do it

mhamendes: You are right, did not see we were using logical values for rendering. This is a big issue with using logical operators for rendering stuff, way too common of a mistake when using non boolean values. I would rather always using `condigiton ? component : null` when conditionally rendering components because you avoid falling on this trap.

src/components/app/pageEvents/components/eventDialogContent.tsx
twistershark: Updated!

src/components/app/pageEvents/components/stateEventsDialogContent.tsx
twistershark: Ok!

src/inngest/functions/eventNotification/index.ts
twistershark: This function in specific get all events created in builder.io, then filters them in two groups and then, for each event, get all records of notifications for that specific event in our DB and then calls another inngest function to send the SMSs. 

Thinking about a real life scenario here, I don't think that we will have so many events happening on the same day to create a bottleneck here and even so, the bulk SMS function is what could take longer, but we already have logic there to handle with batches when needed and so on.

As a bunch of our inngest functions only have one step and the only async calls that we use in this function is getting builder.io data that happens in less than a second and querying the database that won't take more than a few seconds because I added an index on the eventSlug and eventState that are the two informations that we use to query data, I don't think that this scenario will happen, at least for now.

src/inngest/functions/eventNotification/logic.ts
twistershark: As we only create a notification record for each user / event, this prob won't be an issue when we compare with our current website traffic. For this to be an issue, we would need to have 100k users creating the notifications alert for an specific event. We should monitor this, but for now I would just ignore it

src/inngest/functions/eventNotification/index.ts
mhamendes: Oh, in that case everything should be fine, I thought you were sending the notification sms here, if that is not the case, it really should not be an issue.

mhamendes (dismissed):

src/components/app/pageEvents/components/stateEventsDialogContent.tsx
```diff
@@ -14,10 +16,12 @@ interface StateEventsDialogProps {
...
 export function StateEventsDialogContent({ state, events }: StateEventsDialogProps) {
+  usePreventOverscroll()
+
   const parsedState = state.toUpperCase() as keyof typeof US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP
-  const stateEvents = events?.filter(
-    event => event.data.state?.toLowerCase() === state.toLowerCase(),
-  )
+  const stateEvents = events
+    ? events?.filter(event => event.data.state?.toLowerCase() === state.toLowerCase())
+    : []
```
mhamendes, line 24: This should do the trick.

```suggestion
  const stateEvents = events?.filter(event => event.data.state?.toLowerCase() === state.toLowerCase()) ?? []
```

src/components/app/pageEvents/components/eventDialogPhoneNumber.tsx
```diff
@@ -0,0 +1,87 @@
...
+              payload => actionUpdateUserHasOptedInToSMS({ phoneNumber: payload.phoneNumber }),
+            )
+
+            if (result.status === 'success') {
+              await onSuccess()
+            }
+          }, trackFormSubmissionSyncErrors('EventDialogPhoneNumber'))}
+        >
+          <div className="flex w-full items-start gap-4">
+            <div className="flex w-[84px] items-center justify-center rounded-md border border-input bg-background px-2 py-[15px] focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 ">
```
lucasrmp-cb, line 49: same with `w-[84px]` and `py-[15px]`, lets use close theme values

src/components/app/pageEvents/components/eventDialogPhoneNumber.tsx
```diff
@@ -0,0 +1,87 @@
...
+            />
+          </div>
+
+          <Button className="w-full" disabled={form.formState.isSubmitting} type="submit">
+            Get updates
+          </Button>
+        </form>
+      </Form>
+
+      <p className="text-center font-mono text-[10px] text-muted-foreground">
```
lucasrmp-cb, line 78: same for `text-[10px]`

src/components/app/pageEvents/components/eventsIntro.tsx
```diff
@@ -1,12 +1,15 @@
...
 
+import { PageSubTitle } from '@/components/ui/pageSubTitle'
+import { PageTitle } from '@/components/ui/pageTitleText'
+
 export function EventsIntro() {
   return (
     <section className="flex flex-col items-center gap-10">
       <div className="flex flex-col gap-4 lg:gap-6">
-        <h3 className="text-bold text-center font-sans text-xl text-foreground lg:text-[2rem]">
+        <PageTitle className="text-center font-sans text-xl text-foreground lg:text-[2rem]">
```
lucasrmp-cb, line 10: same for `text-[2rem]`, lets find a default size that works here

mhamendes (dismissed):

src/inngest/functions/eventNotification/index.ts
lucasrmp-cb: Got why the timeout is not an issue, but the step is unnecessary, why have this step at all and not just calling the function normally?

lucasrmp-cb (changes requested):

src/components/app/pageEvents/components/allUpcomingEvents.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }, [isMobile])
+
+  return (
+    <section className="flex w-full flex-col items-center gap-4 lg:gap-6">
+      <PageSubTitle className="text-bold font-sans text-xl text-foreground">
+        All upcoming events
+      </PageSubTitle>
+      {!isMobile && (
+        <div className="flex items-center justify-center gap-4 rounded-[8px] bg-backgroundAlternate px-4 py-3">
```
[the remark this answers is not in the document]
twistershark: Removed

src/components/app/pageEvents/components/eventDialog.tsx
```diff
@@ -5,20 +5,27 @@ import { ReactNode } from 'react'
...
 
-export function EventDialog({ event, trigger }: EventDialogProps) {
+export function EventDialog({ event, trigger, triggerClassName }: EventDialogProps) {
   const dialogProps = useDialog({ analytics: 'Event Details Dialog' })
 
   return (
     <Dialog {...dialogProps}>
-      <DialogTrigger className="flex w-full justify-center">{trigger}</DialogTrigger>
-      <DialogContent a11yTitle={`State ${event.state} Events`} className="max-w-[578px]">
+      <DialogTrigger className={triggerClassName ?? 'flex w-full justify-center'}>
```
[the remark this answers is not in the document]
twistershark: The idea is to use triggerClassName or the other, but not both at the same time

src/components/app/pageEvents/components/eventDialog.tsx
```diff
@@ -5,20 +5,27 @@ import { ReactNode } from 'react'
...
+        {trigger}
+      </DialogTrigger>
+      <DialogContent
+        a11yTitle={`State ${event.state} Events`}
+        className="max-w-[578px]"
```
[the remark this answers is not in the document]
twistershark: Updated

src/components/app/pageEvents/components/eventDialogContent.tsx
```diff
@@ -1,121 +1,220 @@
...
+      eventSlug: event.slug,
+      eventState: event.state,
+      shouldReceiveNotifications,
+    }
+
+    const result = await triggerServerActionForForm(
+      {
+        formName: 'RSVP Event',
+        onError: (_, error) => {
+          toast.error(error.message, {
```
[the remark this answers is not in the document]
twistershark: Updated

src/components/app/pageEvents/components/eventDialogPhoneNumber.tsx
twistershark: Changed the w-84px, but i need the padding vertical to be specifically 15px. Can't change that

src/components/app/pageEvents/components/eventDialogPhoneNumber.tsx
twistershark: Updated

src/components/app/pageEvents/components/eventsIntro.tsx
twistershark: Updated

src/components/app/pageEvents/components/eventDialogContent.tsx
```diff
@@ -1,121 +1,220 @@
...
+  const hasPhoneInformation = !!user?.phoneNumber
+  const [isCreatingRsvpEventAction, setIsCreatingRsvpEventAction] = useState(false)
+
   const sectionProps = useSections({
     sections: Object.values(SectionNames),
     initialSectionId: SectionNames.EVENT_INFO,
     analyticsName: 'Event Details Dialog',
   })
 
+  async function handleCreateRsvpAction({
```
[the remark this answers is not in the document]
twistershark: Updated

src/components/app/pageEvents/components/allUpcomingEvents.tsx
```diff
@@ -0,0 +1,103 @@
...
+          <button onClick={() => setDisplayOption('map')}>
+            <MapOptionSVG isActive={displayOption === 'map'} />
+          </button>
+          <div className="h-[2rem] w-[1px] bg-[#D8DCE2]" />
```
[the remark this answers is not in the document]
twistershark: updated

lucasrmp-cb (approved):

src/components/app/pageEvents/components/stateEventsDialog.tsx
```diff
@@ -1,27 +1,33 @@
...
   events?: SWCEvents
+  isOpen: boolean
+  setIsOpen: (open: boolean) => void
 }
 
-export function StateEventsDialog({ state, events }: StateEventsDialogProps) {
+export function StateEventsDialog({ state, events, isOpen, setIsOpen }: StateEventsDialogProps) {
   const dialogProps = useDialog({
-    analytics: `${US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP[state.toUpperCase() as keyof typeof US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP]} Events Dialog`,
+    analytics: `${US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP[state?.toUpperCase() as keyof typeof US_MAIN_STATE_CODE_TO_DISPLAY_NAME_MAP]} Events Dialog`,
```
mhamendes, line 20: If state is null or undefined this will break because null and undefined cannot index a object.

twistershark: This won't ever happen because on the deeplink page we check if the state is present and valid, and we parse all builder.io information as well, so if the state is null|undefined(it is a required field on builder.io) it would just not show on the events page

mhamendes (approved):

lucasrmp-cb (dismissed):

mhamendes (approved):

olavoparno (approved):

olavoparno (dismissed):

lucasrmp-cb (dismissed):

mhamendes (approved):

lucasrmp-cb (approved):

PR #1179, opened 2024-08-13
fix: move campaign name to user communication

LucasdRossi:
## What changed? Why?

I moved the `campaignName` column from `UserCommunicationJourney` to `UserCommunication`.

Here's why: Each user only has one `UserCommunicationJourney` per type (like `BULK_SMS`). Since all bulk communications are tracked in `UserCommunication`, having `campaignName` in `UserCommunicationJourney` would limit us to just one bulk campaign, which isn’t accurate. By moving `campaignName` to `UserCommunication`, we can track multiple bulk campaigns.

## PlanetScale deploy request

[#47](https://app.planetscale.com/stand-with-crypto/swc-web/deploy-requests/47)

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -451,7 +451,6 @@ model UserCommunicationJourney {
   journeyType        UserCommunicationJourneyType @map("journey_type")
   datetimeCreated    DateTime                     @default(now()) @map("datetime_created")
   userCommunications UserCommunication[]
-  campaignName       String?                      @map("campaign_name")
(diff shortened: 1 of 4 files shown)


LucasdRossi:
For bulk communication we'll create more then one `UserCommunicationJourney`, so it makes sense to keep the `campaignName` as it is

PR #1186, opened 2024-08-13
Fix nft mint action

eduardopicolo-cb:
## What changed? Why?

- Added `useWaitForReceipt` hook before invoking the server action. This should ensure that `thirdwebBaseRPCClient.getTransaction()` does not throws an TransactionNotFound error;
- `thirdwebBaseRPCClient` now uses `viemAdapter` from Thirdweb;
- Added Success section to the flow;

## Notes to reviewers

Easter egg from thirdweb docs 
image

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/userActionFormNFTMint/index.tsx
```diff
@@ -83,14 +85,21 @@ export function UserActionFormNFTMint({ trackMount }: { trackMount?: boolean })
...
+          : { debug: true, transactionHash: null, ...sectionProps }
         return <UserActionFormNFTMintTransactionWatch {...params} />
       }
 
       return <UserActionFormNFTMintTransactionWatchSkeleton />
     }
 
+    case UserActionFormNFTMintSectionNames.SUCCESS:
+      return (
+        <UserActionFormSuccessScreen onClose={() => {}}>
```
lucasrmp-cb, line 98: Isn't this gonna cause a `Done` button without action on the success screen?

src/components/app/userActionFormNFTMint/sections/transactionWatch.tsx
```diff
@@ -39,63 +40,63 @@ export type UserActionFormNFTMintTransactionWatchProps =
...
+        if (result.status === 'success') goToSection(UserActionFormNFTMintSectionNames.SUCCESS)
+      })
+    } catch (err) {
+      Sentry.captureException(err, { tags: { domain: 'nftMint/transactionWatch' } })
+    }
+  }, [goToSection, transactionHash])
+
+  useEffect(() => {
+    if (receipt?.status === 'success') createAction()
+  }, [createAction, receipt])
```
lucasrmp-cb, line 94: Can we have some logic ensuring this will only happen once? I'm afraid the receipt updates for any reason after the status is success and we end up triggering this twice

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

PR #1211, opened 2024-08-15
Events page fast follows

twistershark:
## What changed? Why?

Some changes and fixes on the Events page
## Notes to reviewers

Has to change target from `es5` to `es6` on tsconfig.json to be able to use .matchAll to iterate over all matches in the address string to get the valid US state code.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "target": "es5",
+    "target": "ES6",
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "target": "es5",
+    "target": "ES6",
```
olavoparno, line 4: what happened here?

olavoparno (approved):

tsconfig.json
twistershark: I described why I did it in "notes to reviewers" 😉 it's related to a string method named matchAll

mhamendes (approved):

Issue #1062, opened 2024-07-12
Add user identification to Sentry logs on the server side of NextJS


lucasrmp-cb:

olavoparno:
- For API routes we got problems when statically rendering as cookies are not present, ex.: when declaring `export const dynamic = 'error'`.
- Actions are fine fetching cookies on the go, so we can set sessionId.
- Inngest functions we may not have information on who called them as it depends on the usage so we might as well keep sending info manually.

here's a quick summary @lucasrmp-cb

PR #1237, opened 2024-08-19
Feat/add server user identification

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1062
 
fixes 

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/serverWrappers/withUserSession.ts
```diff
@@ -0,0 +1,22 @@
+import * as Sentry from '@sentry/nextjs'
+import { cookies } from 'next/headers'
+import { NextRequest } from 'next/server'
+
+import { USER_SESSION_ID_COOKIE_NAME } from '@/utils/shared/userSessionId'
+
+type HandlerFunction = (request: NextRequest, ...args: any[]) => Promise<Response>
+
+export function withUserSession(fn: HandlerFunction): HandlerFunction {
```
lucasrmp-cb, line 9: The name of this doesn't make a lot of sense, my understanding is that is I'd call something with this wrapping it would include/require the user session to the payload somehow, and also it isn't clear this should only be used with api routes.

A better name for this would be `withRouteMiddleware`, keeping the same pattern as `withServerActionMiddleware`

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

PR #1239, opened 2024-08-19
Fix invalid address logs

eduardopicolo-cb:
closes 

fixes 

## What changed? Why?

Added a flag to avoid sending errors to Sentry. When a client opens the nft mint modal without being authenticated first, the `useGasFee` hook would fail due to not having any wallet address. 

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionFormNFTMint/useCheckoutController.ts
+++ b/src/components/app/userActionFormNFTMint/useCheckoutController.ts
@@ -71,9 +71,11 @@ function useGasFee(quantity: number) {
   const { user } = useThirdwebAuthUser()
   const account = user?.address
 
+  const shouldFetch = contract && account
+
   return useSWR(
-    contract ? { contract } : null,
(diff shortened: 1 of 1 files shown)


mhamendes (dismissed):

src/components/app/userActionFormNFTMint/useCheckoutController.ts
```diff
@@ -71,9 +71,11 @@ function useGasFee(quantity: number) {
   const { user } = useThirdwebAuthUser()
   const account = user?.address
```
twistershark, line 72: Can we change this var name from account to address?

It would also be good to add the address in the extra sentry object so that we can track which address was used when some error happens in this flow.

`extra: {
            quantity,
            contract,
            user,
            address
          },`

twistershark (approved):

PR #1241, opened 2024-08-19
Creates new confirm registration form

twistershark:
## What changed? Why?

This PR adds a new confirm registration step for  users who have registered to vote because it's important for these folks to check their registration because they might have had an issue come up, their address might be out of date, etc

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionFormVoterRegistration/sections/voterRegistrationForm.tsx
+++ b/src/components/app/userActionFormVoterRegistration/sections/voterRegistrationForm.tsx
@@ -37,6 +37,12 @@ const COPY = {
     step2: 'Check voter registration',
     step2Cta: 'Check',
   },
+  registrationConfirm: {
+    title: 'Awesome, thank you for being registered',
+    subtitle: "Let's double check your information to make sure you have no issues voting.",
+    step2: 'Check voter registration',
+    step2Cta: 'Check',
+  },
 } as const
 
 const WY_DISCLAIMER =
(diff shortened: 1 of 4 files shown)


src/components/app/userActionFormVoterRegistration/sections/voterRegistrationForm.tsx
```diff
@@ -87,22 +93,25 @@ function disclaimer(stateCode: USStateCode | undefined) {
...
-    [checkRegistration],
-  )
+  const { title, subtitle, step2, step2Cta } = useMemo(() => {
+    if (registrationConfirm) return COPY.registrationConfirm
+
+    return checkRegistration ? COPY.checkRegistration : COPY.register
+  }, [checkRegistration, registrationConfirm])
 
   const link = stateCode
     ? REGISTRATION_URLS_BY_STATE[stateCode][
```
eduardopicolo-cb, line 117: Maybe this link should also point to `checkRegistrationUrl` when `registrationConfirm` === true

eduardopicolo-cb (changes requested):

src/components/app/userActionFormVoterRegistration/sections/voterRegistrationForm.tsx
twistershark: it's actually the same. we just wanted a new copy

eduardopicolo-cb (approved):

LucasdRossi (approved):

PR #1254, opened 2024-08-21
Removes unnecessary code and libs

twistershark:
## What changed? Why?

This PR removes dead code(with some exceptions like: unused ui components, intl and a few others), removes unnecessary exports and removes unused libs.

With this, we expect to improve bundle size, project organization and possibly improve hot reload timings because of this statement:  `The file you're editing might have other exports in addition to a React component.` in [Hot Reload Doc
](https://nextjs.org/docs/architecture/fast-refresh#limitations).

## Notes to reviewers

All of the analysis around these unnecessary files were done using this awesome lib: https://knip.dev/

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


olavoparno (dismissed):

src/components/ui/form.tsx
```diff
@@ -211,5 +211,4 @@ export {
   FormItemSkeleton,
   FormLabel,
   FormSuccessMessage,
-  useFormField,
```
lucasrmp-cb, line 214: This is supposed to be an abstracted internal lib, I think we should leave this so in case we need this hook in the future it's clear that it should come from here and not from react-hook-form

lucasrmp-cb (changes requested):

src/components/ui/form.tsx
twistershark: Updated!

src/utils/shared/sharedAnalytics.ts
```diff
@@ -1,35 +1,17 @@
...
-  focus = 'Focus',
-  hover = 'Hover',
-  select = 'Select',
-  move = 'Move',
-  render = 'Render',
-  scroll = 'Scroll',
   view = 'View',
-  search = 'Search',
-  keyPress = 'Key Press',
 }
```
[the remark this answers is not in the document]
twistershark: Updated!

lucasrmp-cb (approved):

PR #1275, opened 2024-08-23
Fix: Missing Opt In Action bug

twistershark:
closes [#1263](https://github.com/Stand-With-Crypto/swc-web/issues/1263)

## What changed? Why?

This PR fixes some bugs caused by users being logged in without the opt in action when signing in using wallet.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
+++ b/src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
@@ -46,10 +46,10 @@ export default function UserActionOptInSWCDeepLink() {
   }, [destination, router, urls, queryString])
 
   React.useEffect(() => {
-    if (session.isLoggedIn) {
+    if (session.isLoggedIn && session.hasOptInUserAction) {
       handleRedirectOnLogin()
     }
-  }, [session.isLoggedIn, handleRedirectOnLogin])
+  }, [session.isLoggedIn, session.hasOptInUserAction, handleRedirectOnLogin])
 
   if (session.isLoading || session.isLoggedIn) return null
 
--- a/src/app/[locale]/profile/page.tsx
+++ b/src/app/[locale]/profile/page.tsx
@@ -37,7 +37,7 @@ export default async function Profile({ params, searchParams }: Props) {
 
     return (
       <RedirectToSignUpComponent
-        callbackDestination={value === 'true' ? 'updateProfile' : null}
+        callbackDestination={value === 'true' ? 'updateProfile' : 'home'}
         locale={locale}
       />
     )
--- a/src/app/[locale]/profile/page.tsx
+++ b/src/app/[locale]/profile/page.tsx
@@ -1,3 +1,4 @@
+import { UserActionType } from '@prisma/client'
 import { Metadata } from 'next'
 
 import { PageUserProfile } from '@/components/app/pageUserProfile'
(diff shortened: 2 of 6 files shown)


src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx
Comment (twistershark) [inline src/app/[locale]/(homepageDialogDeeplink)/action/sign-up/page.tsx:1]: This is a safe guard to prevent redirecting to user profile if session is returning that the user isLoggedIn(when logged using wallet) but the optInAction doesn't exist.

This won't be an issue anymore because of this [line](https://github.com/Stand-With-Crypto/swc-web/blob/43e5d8291a6932ad0f1f4eff7927f5bc61245c63/src/utils/server/thirdweb/onLogin.ts#L89), but if for whatever reason this situation happens again in the future, it won't be infinitely looping between profile and sign up page

src/app/[locale]/profile/page.tsx
Comment (twistershark) [inline src/app/[locale]/profile/page.tsx:1]: This is a safe guard to prevent showing the user profile page if session is returning that the user isLoggedIn(when logged using wallet) but the optInAction doesn't exist.

This won't be an issue anymore because of this [line](https://github.com/Stand-With-Crypto/swc-web/blob/43e5d8291a6932ad0f1f4eff7927f5bc61245c63/src/utils/server/thirdweb/onLogin.ts#L89), but if for whatever reason this situation happens again in the future, it won't be infinitely looping between profile and sign up page

src/hooks/useSession.ts
```diff
@@ -16,33 +16,32 @@ export function useSession() {
...
   const handleLogoutSuccess = React.useCallback(() => {
     Cookies.set(USER_SESSION_ID_COOKIE_NAME, generateUserSessionId())
 
-    if (pathname === internalUrls.profile()) {
-      window.location.replace(internalUrls.home())
-    } else {
-      window.location.reload()
-    }
-  }, [internalUrls, pathname])
+    window.location.replace(internalUrls.home())
```
twistershark, line 43: This is a fix that was reverted yesterday without needing. This will clear all query params when the user logout. This will prevent the userId and sessionId from being in query params, causing the user to be automatically logged in after logout

src/utils/server/thirdweb/onLogin.ts
```diff
@@ -86,6 +86,7 @@ export async function login(payload: VerifyLoginPayloadParams) {
     },
     where: {
       userCryptoAddresses: { some: { cryptoAddress, hasBeenVerifiedViaAuth: true } },
+      userActions: { some: { actionType: UserActionType.OPT_IN } },
```
twistershark, line 89: This was the main issue around this bug. We were not checking if the opt in action existed, causing a false positive here returning that the user existed based on wallet address conditions, but with no opt in action.

src/hooks/useSession.ts
```diff
@@ -16,33 +16,31 @@ export function useSession() {
   const { session: thirdwebSession } = useThirdwebSession()
 
   const isLoading = fullProfileRequest.isLoading
+  const user = fullProfileRequest.data?.user
+  const hasOptInUserAction =
+    user?.userActions?.some(userAction => userAction.actionType === UserActionType.OPT_IN) ?? false
```
olavoparno, line 21: I think Array.some already returns false if no match is found, right?

olavoparno (approved):

src/app/api/identified-user/full-profile-info/route.ts
```diff
@@ -16,6 +17,24 @@ async function apiResponseForUserFullProfileInfo() {
         primaryUserCryptoAddress: true,
         primaryUserEmailAddress: true,
         address: true,
+        userActions: {
```
travisbloom-cb, line 20: why do you need all user actions to be fetched here?

travisbloom-cb (commented):
@twistershark where were we not creating a user action on sign in and where in the PR is this addressed?

travisbloom-cb (approved):
JK just saw. Lets make sure we QA all auth flows:
- redirect with sessionId
- created without opt in
- login with opt in

src/app/api/identified-user/full-profile-info/route.ts
twistershark: Because I'm reusing the `getSensitiveDataClientUserAction` function that needs all of them because of the type of record. I could change it to only handle the opt in case in this code, but as the name of the function is user full profile info, i thought that having all information would make more sense

mhamendes (approved):

olavoparno (approved):

PR #1310, opened 2024-08-28
feat: add custom hook to reload due to inactivity

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1283

fixes 

## What changed? Why?

lazy loaded modules are becoming stale and therefore no longer available leading to errors

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/hooks/useReloadDueToInactivity.ts
+++ b/src/hooks/useReloadDueToInactivity.ts
@@ -1,0 +1,35 @@
+import { useEffect, useRef } from 'react'
+
+export function useReloadDueToInactivity({ timeInMinutes }: { timeInMinutes: number }) {
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null)
+
+  const reloadPage = () => {
+    if (typeof window !== 'undefined') {
+      window.location.reload()
+    }
+  }
+
+  const resetTimer = () => {
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current)
+    }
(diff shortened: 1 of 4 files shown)


LucasdRossi (approved):

src/hooks/useReloadDueToInactivity.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    resetTimer()
+
+    return () => {
+      if (timeoutRef.current) {
+        clearTimeout(timeoutRef.current)
+      }
+      events.forEach(event => window.removeEventListener(event, resetTimer))
+    }
+  }, [resetTimer, timeInMinutes])
```
lucasrmp-cb, line 32: We could refactor this to use the debounce function from lodash, it removes the need to handle low level timeouts like that

```suggestion
  const reloadPage = useCallback(() => {
    if (typeof window !== 'undefined') {
      window.location.reload()
    }
  }, [])

  const debouncedReloadPage = useMemo(
    () => debounce(reloadPage, timeInMinutes * 60 * 1000),
    [timeInMinutes]
  );

  useEffect(() => {
    const events = ["mousemove", "keydown", "click", "scroll", "touchstart"];

    events.forEach((event) => window.addEventListener(event, debouncedReloadPage));

    return () => {
      events.forEach((event) => window.removeEventListener(event, debouncedReloadPage));
    };
  }, [debouncedReloadPage]);
```

lucasrmp-cb (changes requested):
Comments above

PR #1316, opened 2024-08-28
Voter guide page

eduardopicolo-cb:
closes 

fixes 

## What changed? Why?

image

image

## Notes to reviewers

Added `shouldBypassAuth` to the pledge action to allow unauthed users to complete the first step.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/pageVoterGuide/keyRacesForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }, [form, mutate, onViewKeyRacesActionSuccess, router])
+
+  const scriptStatus = useGoogleMapsScript()
+
+  useEffect(() => {
+    if (address?.description && racesByAddressRequest.data && scriptStatus === 'ready') {
+      void createViewKeyRacesAction()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
lucasrmp-cb, line 150: Not a big fan of using this for these kinds of cases. I usually recommend creating a ref of the dependency you don't want to trigger and using it inside the effect, that way you don't remove the check on all the other dependencies that should in fact be in the array.

Something like this should do the trick:
```typescriptreact
const createViewKeyRacesActionRef = React.useRef(createViewKeyRacesAction)
createViewKeyRacesActionRef.current = createViewKeyRacesAction

React.useEffect(() => {
  void createViewKeyRacesActionRef.current()
}, [])
```

src/components/app/pageVoterGuide/voterJourneyStepCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+            height={CHECKMARK_SIZE}
+            src={'/misc/uncheckedCircle.svg'}
+            width={CHECKMARK_SIZE}
+          />
+        )
+      case 'unknown':
+      default:
+        return null
+    }
+  }
```
lucasrmp-cb, line 43: We shouldn't use this pattern of rendering due to problems it can cause with component life cycle in some cases. Let's refactor this so this image is a separate component in the same file

src/components/app/userActionFormVotingInformationResearched/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const ANALYTICS_NAME_USER_ACTION_FORM_VOTING_INFORMATION_RESEARCHED =
+  'User Action Form Voting Information Researched'
+
+export enum SECTIONS_NAMES {
+  LOADING = 'Loading',
+  ADDRESS = 'Address',
+  SUCCESS = 'Success',
+}
```
lucasrmp-cb, line 8: Let's keep the pattern for enum naming, it should be pascal case and all the values in constant case.
```suggestion
export enum SectionsNames {
  LOADING = 'Loading',
  ADDRESS = 'Address',
  SUCCESS = 'Success',
}
```

lucasrmp-cb: Gonna add it to eslint

src/components/app/userActionFormVotingInformationResearched/sections/address.tsx
```diff
@@ -0,0 +1,185 @@
...
+  VotingInformationResearchedFormValues,
+} from './formConfig'
+
+export interface AddressProps {
+  onSuccess: (address: z.infer<typeof zodAddress>) => void
+  onSubmit?: () => void
+  initialValues?: Partial<VotingInformationResearchedFormValues>
+}
+
+export const Address = (props: AddressProps) => {
```
lucasrmp-cb, line 41: This should use the `function` to define
https://github.com/Stand-With-Crypto/swc-web/blob/645b0e8cf20a127293e1987a2514219615433169/docs/Coding%20Conventions.md#L27

src/components/app/userActionFormVotingInformationResearched/sections/address.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const router = useRouter()
+  const searchParams = useSearchParams()
+
+  const form = useForm<VotingInformationResearchedFormValues>({
+    defaultValues: initialValues,
+    resolver: zodResolver(votingInformationResearchedFormValidationSchema),
+  })
+
+  const addressField = form.watch('address')
```
lucasrmp-cb, line 52: Let's use `useWatch` here

src/components/app/userActionFormVotingInformationResearched/sections/address.tsx
```diff
@@ -0,0 +1,185 @@
...
+            ...(address ? convertAddressToAnalyticsProperties(address) : {}),
+            'Campaign Name': formValues.campaignName,
+            'User Action Type': UserActionType.VOTING_INFORMATION_RESEARCHED,
+            'Subscribed to notifications': formValues.shouldReceiveNotifications,
+          },
+          payload: { ...formValues, address },
+          onError: (_, e) => {
+            form.setError('root', {
+              message: e.message,
+            })
```
lucasrmp-cb, line 106: You are already toasting a generic message, and we can't guarantee `e.message` is a human readable message. Let's not show this message to the end user

src/components/app/userActionFormVotingInformationResearched/sections/address.tsx
```diff
@@ -0,0 +1,185 @@
...
+        </div>
+
+        <Button
+          className="mx-auto mt-auto"
+          disabled={form.formState.isSubmitting || !addressField?.place_id}
+          form="view-key-races-form"
+          size="lg"
+          type="submit"
+        >
+          Check your voting information
```
lucasrmp-cb, line 180: In other forms we usually show a `Loading...` message when we're submitting (not the prettiest, but effective), let's add it here too
```suggestion
          {form.formState.isSubmitting ? 'Loading...' : 'Check your voting information'}
```

lucasrmp-cb (changes requested):

src/actions/actionCreateUserActionVoterAttestation.ts
```diff
@@ -67,20 +74,25 @@ async function _actionCreateUserActionVoterAttestation(input: CreateActionVoterA
...
   const sessionId = getUserSessionId()
 
-  const authUser = await appRouterGetAuthUser()
-  if (!authUser) {
+  const userMatch = await getMaybeUserAndMethodOfMatch({
+    prisma: {
+      include: { primaryUserCryptoAddress: true, address: true },
+    },
+  })
+  const shouldBypassAuth = validatedInput.data.shouldBypassAuth
```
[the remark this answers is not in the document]
eduardopicolo-cb: This is necessary to reduce friction for the voter guide. We need to allow unauthenticated users to complete the first step

lucasrmp-cb (approved):

src/actions/actionCreateUserActionVoterAttestation.ts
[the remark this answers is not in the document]
lucasrmp-cb: Got this, I'm saying since there's a use case where this route can be called without authentication we can remove it all together and only authenticate on the client when we want to have it

src/components/app/pageVoterGuide/keyRacesForm.tsx
```diff
@@ -0,0 +1,233 @@
...
+  toastGenericError,
+} from '@/utils/web/toastUtils'
+
+interface KeyRacesFormProps {
+  onSubmit?: (formData: VoterGuideFormValues) => Promise<void>
+  onViewKeyRacesActionSuccess?: () => void
+  initialValues?: VoterGuideFormValues
+}
+
+export const KeyRacesForm = (props: KeyRacesFormProps) => {
```
twistershark, line 45: nit pick: Based on our Coding Conventions this should be declared with the function keyword

twistershark (approved):

PR #1329, opened 2024-08-29
ABC Debate email campaign

twistershark:
closes #180 

## What changed? Why?

This PR reuses the same codebase used on CNN email activation to create the ABC presidential debate email campaign.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/actions/actionCreateUserActionEmailDebate.ts
```diff
@@ -0,0 +1,306 @@
...
+  address: Address | null
+}
+type Input = z.infer<typeof zodUserActionFormEmailDebateAction>
+
+export const actionCreateUserActionEmailDebate = withServerActionMiddleware(
+  'actionCreateUserActionEmailDebate',
+  _actionCreateUserActionEmailDebate,
+)
+
+async function _actionCreateUserActionEmailDebate(input: Input) {
```
mhamendes, line 59: Do we need a new action for this? The default email campaign is not enough? Seems like both are very similar, no?

src/components/app/userActionFormEmailDebate/skeleton.tsx
```diff
@@ -0,0 +1,73 @@
+import { Button } from '@/components/ui/button'
+import { FormItemSkeleton } from '@/components/ui/form'
+import { Input } from '@/components/ui/input'
+import { LoadingOverlay } from '@/components/ui/loadingOverlay'
+import { PageSubTitle } from '@/components/ui/pageSubTitle'
+import { PageTitle } from '@/components/ui/pageTitleText'
+import { ScrollArea } from '@/components/ui/scroll-area'
+import { Textarea } from '@/components/ui/textarea'
+
+export function UserActionFormEmailDebateSkeleton() {
```
mhamendes, line 10: The skeleton and modal have different sizes, it's weird how the loading behaves.
skeleton: image
modal: 
image

src/actions/actionCreateUserActionEmailDebate.ts
twistershark: As this action will prob be reused in the future, IMO it's better to keep it separated because if the email action logic changes, we wouldn't need to change anything about the email debate and vice versa

src/components/app/userActionFormEmailDebate/skeleton.tsx
twistershark: Updated

olavoparno (approved):

PR #1348, opened 2024-09-04
User action form for voting_information_researched action

eduardopicolo-cb:
[GH-176](https://github.com/Stand-With-Crypto/swc-internal/issues/176)

## What changed? Why?

Adds the `userActionFormVotingInformationResearched` form & dialog to be used in the Voter Guide page

## Notes to reviewers

Added `initOnMount: false` to the `usePlacesAutocomplete` hook to fix the undefined maps error. We are manually initializing the hook once the `GoogleMapScript`  status is `ready`.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/ui/googlePlacesSelect/index.tsx
+++ b/src/components/ui/googlePlacesSelect/index.tsx
@@ -50,6 +50,7 @@ export const GooglePlacesSelect = React.forwardRef<
       language: 'en',
       types: ['street_address', 'premise', 'postal_code', 'subpremise', 'route'],
     },
+    initOnMount: false,
   })
   // For some countries, the route type is required to show addresses options correctly
   // But for US, if we allow users to use an address with the route type,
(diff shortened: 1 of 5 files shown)


olavoparno (approved):

olavoparno:
@eduardopicolo-cb please add the issue on this PR

src/components/ui/googlePlacesSelect/index.tsx
```diff
@@ -50,6 +50,7 @@ export const GooglePlacesSelect = React.forwardRef<
       language: 'en',
       types: ['street_address', 'premise', 'postal_code', 'subpremise', 'route'],
     },
+    initOnMount: false,
```
twistershark, line 53: Do we need to change other places where we use this hook after this change?

twistershark (approved):

src/components/ui/googlePlacesSelect/index.tsx
eduardopicolo-cb: We have to. But going to do this in another PR to isolate the changes, just in case.

It is going to fix this error 
image

PR #1356, opened 2024-09-04
refactor: adjust script to run faster

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-internal/issues/181

fixes 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/inngest/functions/backfillReactivation/index.tsx
```diff
@@ -144,105 +154,126 @@ export const backfillReactivationWithInngest = inngest.createFunction(
...
+            userSessions: true,
+            address: true,
           },
         })
 
-        for (const user of usersWithoutCommunicationJourney) {
-          const messageId = await sendInitialSignUpEmail(user.id)
+        const messageIds = (await sendBatchEmails(
+          usersWithoutCommunicationJourney as User[],
+        )) as string[]
```
lucasrmp-cb, line 167: Lets remove this cast and set the return type of `sendBatchEmails`, you might need to do a casting on there

src/utils/server/email/sendMail.ts
```diff
@@ -44,15 +60,36 @@ export async function sendMail(payload: SendMailPayload) {
...
+  const response = await SendGrid.send(parsedPayload as MailDataRequired, isMultiple)
+
+  if (isMultiple) {
+    return response.map(
+      currentResponse =>
+        (currentResponse as [ClientResponse])[0].headers?.['x-message-id'] as string,
+    )
+  }
+
+  return response[0].headers?.['x-message-id'] as string
```
lucasrmp-cb, line 94: Is this right? `response` is gonna be an array even when `isMultiple === false`?

lucasrmp-cb (changes requested):

src/inngest/functions/backfillReactivation/index.tsx
olavoparno: I forgot to remove from there. sendMail is type aware thus inferring it.

src/utils/server/email/sendMail.ts
olavoparno: It is always an array, it was like this before:

```ts
const [response] = await SendGrid.send()
```

src/inngest/functions/backfillReactivation/index.tsx
```diff
@@ -144,105 +154,126 @@ export const backfillReactivationWithInngest = inngest.createFunction(
...
+      id: true,
+      userId: true,
+    },
+    distinct: ['userId'],
+  })
 
-    return Promise.reject(err)
+  await prismaClient.userCommunication.createMany({
+    data: journeys.map(({ id, userId }) => ({
+      messageId: emailResults.find(result => result.userId === userId)!.messageId,
```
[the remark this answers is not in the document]
olavoparno: right now nothing happens. I will add it into inngest's response

src/utils/server/email/sendMail.ts
```diff
@@ -44,15 +60,36 @@ export async function sendMail(payload: SendMailPayload) {
...
+        },
+        ...payload,
+      }
+
+  const response = await SendGrid.send(parsedPayload as MailDataRequired, isMultiple)
+
+  if (isMultiple) {
+    return response.map(
+      currentResponse =>
+        (currentResponse as [ClientResponse])[0].headers?.['x-message-id'] as string,
```
LucasdRossi, line 90: do we actually need this as string in both returns?

src/utils/server/email/sendMail.ts
```diff
@@ -44,15 +60,36 @@ export async function sendMail(payload: SendMailPayload) {
...
+        from: SENDGRID_SENDER,
+        mailSettings: {
+          sandboxMode: {
+            enable: SENDGRID_SANDBOX_MODE === 'true',
+          },
+        },
+        ...payload,
+      }
+
+  const response = await SendGrid.send(parsedPayload as MailDataRequired, isMultiple)
```
LucasdRossi, line 85: do we need this `as MailDataRequired`?

src/inngest/functions/backfillReactivation/index.tsx
```diff
@@ -144,105 +155,137 @@ export const backfillReactivationWithInngest = inngest.createFunction(
...
+    errors,
+  }
+}
+
+async function sendBatchEmails(users: User[]) {
+  const emailsPayload = users.map(user => {
+    const userSession = user.userSessions?.[0]
+    const completedActionTypes = user.userActions
+      .filter(action => Object.values(EmailActiveActions).includes(action.actionType))
+      .map(action => action.actionType as EmailActiveActions)
```
LucasdRossi, line 266: do we need this `as EmailActiveActions`?

src/utils/server/email/sendMail.ts
olavoparno: no, we actually don't. thanks for spotting it

src/utils/server/email/sendMail.ts
olavoparno: not really, it was old code before the overload interfaces. thanks man

src/inngest/functions/backfillReactivation/index.tsx
```diff
@@ -107,17 +125,10 @@ export const backfillReactivationWithInngest = inngest.createFunction(
         communicationType: CommunicationType
         messageId: string
       }[]
+      errors: any[]
```
LucasdRossi, line 128: If we don't know the type of the error and we're not using it we should type it as **unknown**, so I would use `unknown[]` here instead of `any[]`

src/inngest/functions/backfillReactivation/index.tsx
```diff
@@ -144,105 +155,137 @@ export const backfillReactivationWithInngest = inngest.createFunction(
...
-  const currentSession = userSession
-    ? {
-        userId: userSession.userId,
-        sessionId: userSession.id,
-      }
-    : null
-
-  const ReactivationReminderComponent = ReactivationReminder
+async function persistBatchUserCommunication(emailResults: EmailResult[]) {
+  const errors: any[] = []
```
LucasdRossi, line 199: same as the above comment

src/inngest/functions/backfillReactivation/index.tsx
olavoparno: thanks

LucasdRossi (dismissed):

twistershark (approved):

lucasrmp-cb (approved):

PR #1367, opened 2024-09-05
fix: remove lodash template from addVariablesToMessage

LucasdRossi:
closes 

fixes 

## What changed? Why?

The lodash `template` function was causing timeouts on Vercel, preventing messages from being sent, so I switched to a simpler approach using regex.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/inngest/functions/sms/utils/enqueueMessages.ts
+++ b/src/inngest/functions/sms/utils/enqueueMessages.ts
@@ -1,7 +1,7 @@
 import { UserCommunicationJourneyType } from '@prisma/client'
 import * as Sentry from '@sentry/node'
 import { NonRetriableError } from 'inngest'
-import { template, update } from 'lodash-es'
+import { update } from 'lodash-es'
 
 import { sendSMS, SendSMSError } from '@/utils/server/sms'
 import { optOutUser } from '@/utils/server/sms/actions'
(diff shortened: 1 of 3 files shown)


twistershark (dismissed):

src/inngest/functions/sms/utils/enqueueMessages.ts
```diff
@@ -204,3 +213,16 @@ export function countMessagesAndSegments(payload: EnqueueMessagePayload[]) {
...
+interface Variables {
+  firstName?: string
+  lastName?: string
+  sessionId?: string
+  userId?: string
+}
+
+function addVariablesToMessage(message: string, variables: Variables) {
+  return message.replace(/{{\s*(\w+)\s*}}/g, (_, variable: keyof Variables) =>
+    variables[variable] !== undefined ? (variables[variable] ?? '') : '',
```
olavoparno, line 226: we can directly test against variables[key] instead of testing against undefined. also, prefer asserting with typeof rather than the object of u defined directly like typeof === ‘undefined’

olavoparno (dismissed):

olavoparno (approved):

PR #1384, opened 2024-09-10
Tweaks on events page

### LINKED ISSUE #1387: Events map not showing more than 3 dots for state with more than 3 events
We need to update how we display the events in each state in the events map to represent the correct amount of events on each state.

Based on what our designer suggested, we should show only one circle with the number of events inside of it.

twistershark:
closes #1387

## What changed? Why?

This PR improves on the events page map component by changing markers coordinates.

I had to create a new coords const because the markers where not being correctly centered in a lot of states. I checked every coordinate for every state to make sure that it will render correctly and with good spacing.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Screenshot

image


src/components/app/pageEvents/components/eventsMap.tsx
```diff
@@ -52,29 +51,20 @@ export function EventsMap({ events }: { events: SWCEvents }) {
...
-          id: `${state} ${i}`,
-          name: `${state} ${i}`,
-          coordinates: offsetCoordinates,
-        }
-
-        result.push(marker)
+      const marker: MapMarker = {
+        id: `${state}`,
+        name: `${state}`,
+        coordinates: [coordinates[0], coordinates[1]],
```
lucasrmp-cb, line 59: Looks like the same thing

```suggestion
        id: state,
        name: state,
        coordinates,
```

lucasrmp-cb (commented):
Little refactor, and can I get a better description on why we need to duplicate the coords const for this screen? What's the issue here?

twistershark:
@lucasrmp-cb I had to create a new coords const because the markers where not being correctly centered in a lot of states. I checked every coordinate for every state to make sure that it will render correctly and with good spacing.

olavoparno (dismissed):

lucasrmp-cb:
> @lucasrmp-cb I had to create a new coords const because the markers where not being correctly centered in a lot of states. I checked every coordinate for every state to make sure that it will render correctly and with good spacing.

And why duplicating the coordinates, doesn't it mean that they are off in the activity map too? Also, changing it in the original breaks anything?
And finally, can we add all this context (maybe some screenshots even) to the PR description and create an issue to have this tracked?

twistershark:
> > @lucasrmp-cb I had to create a new coords const because the markers where not being correctly centered in a lot of states. I checked every coordinate for every state to make sure that it will render correctly and with good spacing.
> 
> And why duplicating the coordinates, doesn't it mean that they are off in the activity map too? Also, changing it in the original breaks anything? And finally, can we add all this context (maybe some screenshots even) to the PR description and create an issue to have this tracked?

On the Activity Map, the user doesn't have any interaction with each state, so it doesn't really matter if it not centered on the state perimeter or not. I checked with Olavo and if I understood correctly, the original coordinates points to the capital coordinates on each state, while on the events page I changed it to be more visually pleasing.

lucasrmp-cb (approved):

olavoparno (approved):

PR #1432, opened 2024-09-18
feat: child logger

LucasdRossi:
closes [#125](https://github.com/Stand-With-Crypto/swc-internal/issues/125)

fixes 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/inngest/functions/backfillCongressionalDistrictCronJob/index.ts
+++ b/src/inngest/functions/backfillCongressionalDistrictCronJob/index.ts
@@ -5,7 +5,7 @@ import { inngest } from '@/inngest/inngest'
 import { onScriptFailure } from '@/inngest/onScriptFailure'
 import { prismaClient } from '@/utils/server/prismaClient'
 import { maybeGetCongressionalDistrictFromAddress } from '@/utils/shared/getCongressionalDistrictFromAddress'
-import { getLogger } from '@/utils/shared/logger'
+import { Logger } from '@/utils/shared/logger'
 import { NEXT_PUBLIC_ENVIRONMENT } from '@/utils/shared/sharedEnv'
 
 const BACKFILL_US_CONGRESSIONAL_DISTRICTS_INNGEST_CRON_JOB_FUNCTION_ID =
@@ -22,7 +22,6 @@ const BACKFILL_US_CONGRESSIONAL_DISTRICTS_BATCH_SIZE =
 const MAX_US_CONGRESSIONAL_DISTRICTS_BACKFILL_COUNT =
   Number(process.env.MAX_US_CONGRESSIONAL_DISTRICTS_BACKFILL_COUNT) || 150000 // Our quota is 250000 queries per day
 
-const logger = getLogger('backfillUsCongressionalDistrictsCronJob')
 export const backfillCongressionalDistrictCronJob = inngest.createFunction(
   {
     id: BACKFILL_US_CONGRESSIONAL_DISTRICTS_INNGEST_CRON_JOB_FUNCTION_ID,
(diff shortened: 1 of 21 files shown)


olavoparno (approved):

src/inngest/functions/setPrimaryCryptoAddressOfUser/logic.ts
```diff
@@ -13,11 +11,12 @@ const zodPrimaryCryptoAddressOfUserParameters = object({
 
 export async function setPrimaryCryptoAddressOfUser(
   parameters: z.infer<typeof zodPrimaryCryptoAddressOfUserParameters>,
+  logger?: Logger,
```
lucasrmp-cb, line 14: Same here, if the idea is make `setPrimaryCryptoAddressOfUser` resilient we should just default this value to the return of `getLogger`
```suggestion
  logger: Logger = getLogger('setPrimaryCryptoAddressOfUser'),
```

src/inngest/functions/sms/utils/enqueueMessages.ts
```diff
@@ -35,7 +33,11 @@ export interface EnqueueMessagePayload {
   messages: PayloadMessage[]
 }
 
-export async function enqueueMessages(payload: EnqueueMessagePayload[], attempt = 0) {
+export async function enqueueMessages(
+  payload: EnqueueMessagePayload[],
+  logger?: Logger,
```
lucasrmp-cb, line 38: Same here, and let's merge `logger` and `attempt` in a `options` object, the API is easier to read that way

lucasrmp-cb (changes requested):

src/inngest/functions/backfillCongressionalDistrictCronJob/index.ts
```diff
@@ -94,6 +93,7 @@ async function backfillUsCongressionalDistricts(
     Address,
     'id' | 'formattedDescription' | 'countryCode'
   >[],
+  logger?: Logger,
```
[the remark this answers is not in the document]
LucasdRossi: I thought about not linking the logic of the functions to a specific type in Inngest. There are some functions that separate the logic from the initialization and execution of Inngest:
[src/inngest/functions/backfillNFT
](https://github.com/Stand-With-Crypto/swc-web/tree/main/src/inngest/functions/backfillNFT)
[src/inngest/functions/cleanupPostalCodes ](https://github.com/Stand-With-Crypto/swc-web/tree/main/src/inngest/functions/cleanupPostalCodes)

olavoparno (dismissed):

src/inngest/functions/backfillCongressionalDistrictCronJob/index.ts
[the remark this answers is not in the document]
lucasrmp-cb: You have a few options to suppress or even test console calls in jest, I think we can go towards something like this instead.
image
Maybe we wanna mock logs based on and env or something like that. It serves as a test debug sometimes, I dont think we should change implementation thinking of tests but the other way around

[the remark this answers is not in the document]
lucasrmp-cb: And about not creating inngest dependencies. I get that and agree to some extend, but I'm ok with creating inngest dependencies inside the `inngest` folder

src/inngest/functions/cleanupPostalCodes/logic.ts
```diff
@@ -1,7 +1,9 @@
 import { prismaClient } from '@/utils/server/prismaClient'
-import { Logger } from '@/utils/shared/logger'
+import { getLogger } from '@/utils/shared/logger'
 
-export async function cleanPostalCodes(persist: boolean, logger?: Logger) {
+const defaultLogger = getLogger('cleanPostalCodes')
+
+export async function cleanPostalCodes(persist: boolean, logger = defaultLogger) {
```
lucasrmp-cb, line 6: its still a np of mine (and you don't need to change it), but I don't like the signature call of `cleanPostalCodes(true, logger)`, I'd much prefer `cleanPostalCodes({ persist: true, logger })`, this way I know what `true` means

src/inngest/functions/setPrimaryCryptoAddressOfUser/logic.ts
```diff
@@ -36,7 +38,7 @@ export async function setPrimaryCryptoAddressOfUser(
     throw new Error('a user with the same primaryUserCryptoAddressId already exist')
   }
   if (!persist) {
-    logger?.info('Dry run, exiting')
+    logger.info('Dry run, exiting')
```
lucasrmp-cb, line 41: I know this is not your task, but I think we're missing a return here. Big problem so if you can check and fix

lucasrmp-cb (changes requested):

olavoparno (approved):

lucasrmp-cb (approved):

PR #1437, opened 2024-09-19
Merge verified users

### LINKED ISSUE #1382: Update onLogin logic so that if a user is anonymous, performs a few actions, then logs in, we merge those accounts, even if the logged in account already exists

eduardopicolo-cb:
closes #1382 

fixes 

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/server/thirdweb/onLogin.ts
+++ b/src/utils/server/thirdweb/onLogin.ts
@@ -99,7 +99,6 @@ export async function login(payload: VerifyLoginPayloadParams) {
       existingVerifiedUser,
       cryptoAddress,
       localUser,
-      getUserSessionId: () => _getUserSessionId(),
     })
 
     await Promise.all([
@@ -161,7 +160,7 @@ type ExistingUserLoginParams = {
   existingVerifiedUser: User
   cryptoAddress: string
   localUser: ServerLocalUser | null
-  getUserSessionId: () => string | null
+  // getUserSessionId: () => string | null
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/utils/server/thirdweb/onLogin.ts
```diff
@@ -149,6 +156,67 @@ export async function login(payload: VerifyLoginPayloadParams) {
   cookies().set(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX, jwt)
 }
 
+type ExistingUserLoginParams = {
+  existingVerifiedUser: User
+  cryptoAddress: string
+  localUser: ServerLocalUser | null
+  // getUserSessionId: () => string | null
```
olavoparno, line 163: could this comment be removed?

olavoparno (approved):

src/utils/server/thirdweb/onLogin.ts
```diff
@@ -93,7 +93,14 @@ export async function login(payload: VerifyLoginPayloadParams) {
   })
 
   if (existingVerifiedUser) {
-    log('existing user found')
+    log('existing verified user found')
+
+    await onExistingUserLogin({
```
travisbloom-cb, line 98: We'll want to audit how expensive this query current is in prod. I think its pretty slow :( Wasnt as a big a deal when it was only run for new users but if its run every login it could warrent us figuring out how to optimize

travisbloom-cb: Lets also make sure we update the comment at the top of the file to account for this new use case

olavoparno (approved):

Issue #1434, opened 2024-09-19
Add inngest schemas


olavoparno:
add type inferring into inngest functions 

https://www.inngest.com/docs/reference/client/create#defining-event-payload-types

PR #1438, opened 2024-09-20
Refactor/add inngest schemas

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1434

fixes 

## What changed? Why?

Adds type inference on any inngest method like createFunction and send

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/inngest/functions/initialSignupUserCommunicationJourney/initialSignupUserCommunicationJourney.tsx
```diff
@@ -46,7 +49,7 @@ export const initialSignUpUserCommunicationJourney = inngest.createFunction(
   },
   { event: INITIAL_SIGNUP_USER_COMMUNICATION_JOURNEY_INNGEST_EVENT_NAME },
   async ({ event, step }) => {
-    const payload = await initialSignUpUserCommunicationJourneyPayload
+    const { data: payload } = await initialSignUpUserCommunicationJourneyPayload
       .parseAsync(event.data)
```
LucasdRossi, line 53: Do we need to use `parseAsync` here?

LucasdRossi:
I think we should use PascalCase for defining the event schema types since it's the convention in TypeScript:

```typescript
type BULK_SMS_COMMUNICATION_JOURNEY_INNGEST_EVENT_SCHEMA = {}

type BulkSMSCommunicationJourneyInngestEventSchema = {}
```

LucasdRossi:
If you search for `await inngest.send` in the codebase, you'll find all the instances where we're executing Inngest events. In most of these cases, we're importing the event type, but with this refactor, that's no longer necessary. By simply passing the event name, the payload is already type-checked. I think we should go through these files and remove the payload definitions.

For example, [here](https://github.com/Stand-With-Crypto/swc-web/blob/refactor/add-inngest-schemas/src/bin/smokeTests/capitolCanary/createAdvocateWithInngest.ts#L27) we're defining the payload before calling inngest, but we don't need to do this anymore, we could just do

```typescript
  await inngest.send({
    name: CAPITOL_CANARY_UPSERT_ADVOCATE_INNGEST_EVENT_NAME,
    data: {
    campaignId: getCapitolCanaryCampaignID(CapitolCanaryCampaignName.DEFAULT_MEMBERSHIP),
    user: {
      ...mockedUser,
      address: mockedAddress,
    },
    userEmailAddress: mockedEmailAddress,
    opts: {
      isEmailOptin: true,
    },
    metadata: {
      tags: ['C4 Member', 'Smoke Test User'],
    },
  },
  })
```

and `data` would be type-safe

LucasdRossi (approved):

src/inngest/functions/initialSignupUserCommunicationJourney/initialSignupUserCommunicationJourney.tsx
```diff
@@ -46,7 +49,7 @@ export const initialSignUpUserCommunicationJourney = inngest.createFunction(
   },
   { event: INITIAL_SIGNUP_USER_COMMUNICATION_JOURNEY_INNGEST_EVENT_NAME },
   async ({ event, step }) => {
-    const payload = await initialSignUpUserCommunicationJourneyPayload
+    const { data: payload } = await initialSignUpUserCommunicationJourneyPayload
```
lucasrmp-cb, line 52: This is wrong isn't it?
You changed the zod schema from 
```
{
  userId: string;
  sessionId: string;
  decreaseTimers?: boolean;
}
```

To
```
{
  name: string
  data: {
    userId: string;
    sessionId: string;
    decreaseTimers?: boolean;
  }
}
```

This will fail validation every time

lucasrmp-cb: Let's keep the validation schema as it was and just expand the Schema type 

```
export interface InitialSignupUserCommunicationSchema {
  name: typeof INITIAL_SIGNUP_USER_COMMUNICATION_JOURNEY_INNGEST_EVENT_NAME;
  data: z.infer<
    typeof initialSignUpUserCommunicationJourneyPayload
  >
}
```

src/inngest/types.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { UnstopConfirmationSmsCommunicationJourneyInngestEventSchema } from '@/inngest/functions/sms/unstopConfirmationSMSCommunicationJourney'
+import type { WelcomeSmsCommunicationJourneyInngestEventSchema } from '@/inngest/functions/sms/welcomeSMSCommunicationJourney'
+import type { DeleteUserActionsInngestEventSchema } from '@/inngest/functions/user/deleteUserActions'
+import type { AuditUserBatchEventSchema } from '@/inngest/functions/usersTotalDonationAmountUsd/audit'
+import type {
+  BackfillUsersTotalDonationAmountUsdEventSchema,
+  UpdateUserBatchEventSchema,
+} from '@/inngest/functions/usersTotalDonationAmountUsd/backfill'
+
+type EventTypes =
```
lucasrmp-cb, line 34: Since we have a new step to create inngest functions, let's create a very small section on `docs/Contributing.md` on how to create an inngest function (create the schema, the function itself using `inngest.createFunction` and then add the function to the route file and the schema to this union

src/inngest/functions/sms/utils/enqueueMessages.test.ts
```diff
@@ -82,6 +83,13 @@ jest.mock('@/utils/server/sms', () => ({
   }),
 }))
 
+const mockedLogger: Logger = {
+  debug: jest.fn(),
+  error: jest.fn(),
+  info: jest.fn(),
+  warn: jest.fn(),
+}
```
lucasrmp-cb, line 91: Pretty sure this came from the commits that shouldn't be here but just in case you did it. 
Mocking the logs this way will suppress a bunch of stuff we want to keep showing, let's not do it

lucasrmp-cb (changes requested):
First of all, great job and great idea.

A few things to note:
- We still have changes on the diff from the base branch, to sort this out you can either create a new branch (from main this time) and cherry pick the commits you want from this one or you can manually rebase and drop commits that shouldn't be here.
- This is a np but since this will be a core part of the inngest integration let's change all schemas to use the `interface` keyword, is a code convention we use

olavoparno:
closing PR in favor of https://github.com/Stand-With-Crypto/swc-web/pull/1447

Issue #948, opened 2024-06-06
DTSI analysis UX


travisbloom-cb:
@coinbasemar drafted up some [initial mockups for this UX](https://docs.google.com/document/d/1_Tkm7HncxpK0RZDqdKHIb-GheLL1_VtLrtXw9-00ZkY/edit). There are two major missing components:
- we need a CTA that makes it clear to people they can add their own analysis. We prob want that to open a modal explaining they're leaving the site for dotheysupportit otherwise it'll be a jarring experience
- Can we link to the figma of the original mockup?

PR #1445, opened 2024-09-23
Add dtsi analysis to stances

travisbloom-cb:
closes https://github.com/Stand-With-Crypto/swc-web/issues/948

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/dtsiStanceDetails/index.tsx
+++ b/src/components/app/dtsiStanceDetails/index.tsx
@@ -51,80 +51,86 @@ export function DTSIStanceDetails({ className, ...props }: IStanceDetailsProps) 
     Boolean(stance.analysis.length) || Boolean(stance.additionalAnalysis.length)
   const additionalAnalysisUsers = stance.additionalAnalysis.map(x => x.publicUser)
   return (
-    <article className={cn('rounded-3xl bg-secondary p-4 md:p-6', className)}>
+    <article className={cn('rounded-3xl bg-secondary p-4 sm:p-6', className)}>
       <StanceTypeContent {...props} />
       {stance.stanceType !== DTSI_PersonStanceType.BILL_RELATIONSHIP && (
-        <div className="flex items-center justify-between">
+        <div className="flex flex-col space-y-2 sm:flex-row sm:items-center sm:justify-between">
           <CryptoSupportHighlight className="mt-4 rounded-full py-2" stanceScore={stanceScore} />
           <Dialog analytics={'View DTSI Analysis'}>
             <DialogTrigger asChild>
-              <Button>View Analysis</Button>
+              <Button className="bg-transparent text-primary hover:bg-gray-300 max-sm:hover:underline sm:bg-primary sm:text-primary-foreground sm:hover:bg-primary/80">
+                View Analysis
+              </Button>
             </DialogTrigger>
             <DialogContent a11yTitle="View DTSI Analysis" className="max-w-2xl">
-              <DialogBody className="space-y-8">
-                <div>
-                  <h3 className="text-xl font-bold">Top analysis on stance</h3>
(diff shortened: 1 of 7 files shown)


src/components/app/dtsiStanceDetails/index.tsx
```diff
@@ -43,12 +47,94 @@ function StanceTypeContent({ stance: passedStance, ...props }: IStanceDetailsPro
...
+                                      .map(user => user.displayName),
+                                    `and ${additionalAnalysisUsers.length - 3} other users`,
+                                  ]
+                                : additionalAnalysisUsers.slice(0, 3).map(user => user.displayName),
+                            )}
+                          </div>
+                        </div>
+                      )}
+                    </>
+                  )}
```
lucasrmp-cb: I'm a bit confused in how we got to this UI
image

My understanding from the copy is that we have a main analysis and additional ones. Don't know if this is an error or a preview env inconsistency, but I think `additionalAnalysisUsers` was supposed to be empty here

src/components/app/dtsiUserAvatar.tsx
lucasrmp-cb, line 1: np: This file is a bit out of code conventions, specifically for using function keywords for top level functions and components and interfaces when possible

src/components/app/dtsiUserAvatar.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if (size < 20) {
+    return twNoop('text-sm')
+  }
+  if (size < 30) {
+    return twNoop('text-base')
+  }
+  return twNoop('text-lg')
+}
+
+const Avatar: React.FC<AvatarProps> = ({ size, className, style, ...props }) => {
```
lucasrmp-cb, line 29: Is the idea here that we crop bigger images? SWC shield looks bad IMO
image

src/components/app/dtsiUserAvatar.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if ('src' in props) {
+    return (
+      <div className={cn('relative', className)} style={{ width: size, height: size, ...style }}>
+        <NextImage
+          className={cn('rounded-full')}
+          fill
+          sizes={`${size}px`}
+          style={{ objectFit: 'cover' }}
```
lucasrmp-cb, line 37: np: `NextImage` has a `objectFit` prop, not sure if they do anything else that pass it as a style but just in case

lucasrmp-cb (changes requested):

src/components/app/dtsiStanceDetails/index.tsx
travisbloom-cb: Yes, this was a bug with the DTSI api where additional analysis also include approved analysis. that has been fixed

src/components/app/dtsiUserAvatar.tsx
travisbloom-cb: Good call out, we can update the DTSI image for SWC to add some padding

src/components/app/dtsiUserAvatar.tsx
travisbloom-cb: I dont see in the docs :( How would you adjust?

twistershark (approved):

src/components/app/dtsiUserAvatar.tsx
lucasrmp-cb: Well, I guess not having it in the docs means it's only a shorthand then. But you could do
```suggestion
          objectFit="cover"
```

lucasrmp-cb (dismissed):

lucasrmp-cb (approved):

PR #1448, opened 2024-09-23
refactor: adjust query in two steps so it runs faster

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-internal/issues/150

fixes  https://stand-with-crypto.sentry.io/issues/5585303928/?project=4506490717470720&referrer=github_integration

## What changed? Why?

An attempt to diminish DB joins so it runs faster

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


eduardopicolo-cb (commented):
Can we try to index `primaryUserCryptoAddressId` on the User table and `userId` & `nftMintId` on the UserAction table?

Then query by Id to allow the db to use the index on the foreign key directly: 

```typescript
const userIdsWithCryptoAddress = await prismaClient.user.findMany({
  where: {
    primaryUserCryptoAddressId: { isNot: null }
  },
  select: {
    id: true
  }
}).then(users => users.map(user => user.id))

return await prismaClient.userAction.count({
  where: {
    datetimeCreated: { gte: GO_LIVE_DATE },
    nftMintId: null,
    actionType: { in: actionsWithNFT },
    userId: { in: userIdsWithCryptoAddress }
  }
});
```

olavoparno:
> Can we try to index `primaryUserCryptoAddressId` on the User table and `userId` & `nftMintId` on the UserAction table?
> 
> Then query by Id to allow the db to use the index on the foreign key directly:
> 
> ```ts
> const userIdsWithCryptoAddress = await prismaClient.user.findMany({
>   where: {
>     primaryUserCryptoAddressId: { isNot: null }
>   },
>   select: {
>     id: true
>   }
> }).then(users => users.map(user => user.id))
> 
> return await prismaClient.userAction.count({
>   where: {
>     datetimeCreated: { gte: GO_LIVE_DATE },
>     nftMintId: null,
>     actionType: { in: actionsWithNFT },
>     userId: { in: userIdsWithCryptoAddress }
>   }
> });
> ```

we already have an index for `primaryUserCryptoAddressId` on the User table (because it is an unique ID, I think) and as for the `userId & nftMintId`, we could indeed add and test it out. 

Still, my approach would be much less interventionist and straightforward

eduardopicolo-cb (dismissed):

olavoparno:
@eduardopicolo-cb I've added the index, can you approve again?

eduardopicolo-cb (approved):

PR #1456, opened 2024-09-25
Erase sessionId param from url

eduardopicolo-cb:
## What changed? Why?

Remove `sessionId` & `userId`  params from url. This will not have an impact on client analytics or logout the user.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/[locale]/topLevelClientLogic.tsx
+++ b/src/app/[locale]/topLevelClientLogic.tsx
@@ -12,6 +12,7 @@ import { useReloadDueToInactivity } from '@/hooks/useReloadDueToInactivity'
 import { SupportedLocale } from '@/intl/locales'
 import { AnalyticActionType, AnalyticComponentType } from '@/utils/shared/sharedAnalytics'
 import { thirdwebClient } from '@/utils/shared/thirdwebClient'
+import { getUserIdOnClient } from '@/utils/shared/userId'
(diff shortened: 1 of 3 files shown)


src/app/[locale]/topLevelClientLogic.tsx
```diff
@@ -67,6 +68,16 @@ const InitialOrchestration = () => {
       action: AnalyticActionType.view,
     })
   }, [pathname])
+  const searchParamsSessionId = searchParams?.get('sessionId')
```
Comment (olavoparno) [inline src/app/[locale]/topLevelClientLogic.tsx:71]: if we're not using `searchParamsSessionId` outside of the useEffect could we move it inside?

src/app/[locale]/topLevelClientLogic.tsx
```diff
@@ -37,20 +39,20 @@ const InitialOrchestration = () => {
       Sentry.setUser({ id: sessionId, idType: 'session' })
     }
   }, [])
-  const searchParamsUserId = searchParams?.get('userId')
+  const clientUserId = getUserIdOnClient()
```
Comment (olavoparno) [inline src/app/[locale]/topLevelClientLogic.tsx:42]: same thing here, do we need it outside of useEffect?

olavoparno (approved):

PR #1457, opened 2024-09-25
fix: bulk sms vercel timeout

LucasdRossi:
closes [#234](https://github.com/Stand-With-Crypto/swc-internal/issues/234)

fixes [PROD-SWC-WEB-4ET](https://stand-with-crypto.sentry.io/issues/5691014147/events/5c424ef2701d4c9fb948be7a6a1fadb9/)

## What changed? Why?

This update improves the `bulk-sms` process by making the following changes:

- **Separated `enqueueMessages` Function**: 
  - Moved `enqueueMessages` to its own Inngest function to reduce steps and memory usage. 
  - The previous setup was doing too much at once—fetching variables, queueing messages, and saving to the database—which led to timeouts on Vercel. 
  - In the future, we may run this function in parallel for better performance.

- **Updated Phone Number Fetching**: 
  - We encountered an `output_too_large` error when trying to fetch more than 100k phone numbers. Now, the script fetches 100k phone numbers per step, resolving the issue.
  - We switched to using Prisma's `skip` instead of cursor pagination, which was causing unexpected issues.

- **Refactor `countMessagesAndSegments` Logic**: 
  - The previous implementation caused timeouts because it called `countSegments` for each message. 
  - We now use a hash map to count how many users will receive each message first, and then calculate segments for unique messages. This reduces the number of `countSegments` calls and improves efficiency.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [x] Unit test
- [ ] Functional test


src/inngest/functions/sms/bulkSMSCommunicationJourney.ts
```diff
@@ -92,80 +92,126 @@ export const bulkSMSCommunicationJourney = inngest.createFunction(
...
+                }
+              : {
+                  every: {
+                    journeyType: {
+                      not: UserCommunicationJourneyType.WELCOME_SMS,
+                    },
+                  },
+                },
+          },
+        )
```
olavoparno, line 122: awesome man

src/inngest/functions/sms/enqueueMessages.ts
```diff
@@ -0,0 +1,264 @@
...
+            } else {
+              Sentry.captureException(`sendSMS unexpected Error: ${(error as any)?.message}`, {
+                extra: {
+                  reason: error,
+                },
+                tags: {
+                  domain: 'enqueueMessages',
+                },
+              })
+            }
```
olavoparno, line 169: can this complexity be reduced? maybe abstracting into a separate function with early returns or into an object map?

lucasrmp-cb (approved):

PR #1459, opened 2024-09-26
feat: add more info to email events

lucasrmp-cb:
## What changed? Why?

While building mixpanel dashboards we noticed we didn't have a way to group url by origin, that caused a lot of versions of the same url to be visible, this fixes it and adds a new field to help identify which campaign that click is coming from (redundant field in some cases, but useful in others)

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/api/public/sendgrid/events/route.ts
+++ b/src/app/api/public/sendgrid/events/route.ts
@@ -61,9 +61,10 @@ async function processEventChunk(messageId: string, events: EmailEvent[]) {
       'Sendgrid Message Id': messageId,
       'Sendgrid Event Id': eventEntry.sg_event_id,
       ...(eventEntry.useragent && { 'User Agent': eventEntry.useragent }),
-      ...(eventEntry.url && { Url: eventEntry.url }),
+      ...(eventEntry.url && { Url: new URL(eventEntry.url)?.origin }),
       ...(eventEntry.variant && { Variant: eventEntry.variant }),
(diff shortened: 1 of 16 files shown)


LucasdRossi (approved):

src/app/api/public/sendgrid/events/route.ts
```diff
@@ -61,9 +61,10 @@ async function processEventChunk(messageId: string, events: EmailEvent[]) {
       'Sendgrid Message Id': messageId,
       'Sendgrid Event Id': eventEntry.sg_event_id,
       ...(eventEntry.useragent && { 'User Agent': eventEntry.useragent }),
-      ...(eventEntry.url && { Url: eventEntry.url }),
+      ...(eventEntry.url && { Url: new URL(eventEntry.url)?.origin }),
```
olavoparno, line 64: do we have control over eventEntry.url? It could break here if not

olavoparno (approved):

src/app/api/public/sendgrid/events/route.ts
lucasrmp-cb: According to the sendgrid docs this either will not be in the payload or it will be a valid url
https://www.twilio.com/docs/sendgrid/for-developers/tracking-events/event#json-objects

Issue #1461, opened 2024-09-26
Election night live results


travisbloom-cb:
Figma: https://www.figma.com/design/NfKwbY2nWgWOogRlKhS7bB/Voter-Journey?node-id=1009-482&node-type=canvas&t=G3ACpakIWObctN8m-11

coinbasemar:
We're repurposing our Key Races page to show live election data on election night. We should add the following:

1. A "Live election results" button in the header (remove key races tab in header in favor of this)
2. A Trump vs. Kamala progress bar that shows how many votes each party is receiving
3. A crypto election map that shows the number of pro-crypto candidates being elected in each state, with the ability to click in and view each state's page
4. An address bar for viewing elections in a user's area
5. Updated cards for the 13 critical elections that we show on the key races page
6. An upsell to complete the "I Voted" flow

This work also includes updating the state and race level pages with relevant changes.

Link to Figma (WIP): https://www.figma.com/design/NfKwbY2nWgWOogRlKhS7bB/Voter-Journey?node-id=1009-482

travisbloom-cb:
- Add report that says any matching/lack of matching anomalies that get returned from decision desk/DTSI

PR #1466, opened 2024-09-28
feat: add decisionDesk basic services

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1461

fixes 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


eduardopicolo-cb (approved):

src/data/decisionDesk/services.ts
```diff
@@ -0,0 +1,166 @@
...
+    endpointURL.search = URLParams.toString()
+  }
+
+  const bearerToken = await getBearerToken()
+
+  if (!bearerToken) {
+    throw new Error('Bearer key not found')
+  }
+
+  console.log('endpointURL', endpointURL)
```
eduardopicolo-cb, line 107: console.log

eduardopicolo-cb (changes requested):

src/data/decisionDesk/services.ts
olavoparno: removed thanks

src/data/decisionDesk/services.ts
```diff
@@ -0,0 +1,263 @@
...
+      ),
+    {
+      retries: 1,
+      minTimeout: 4000,
+    },
+  )
+
+  logger.debug(`fetchRacesData returned with status ${response.status}`)
+
+  const json = (await response.json()) as { data: GetRacesResponse } | { errors: any[] }
```
eduardopicolo-cb, line 162: The type cast is not matching the api return. It should be `GetRacesResponse` instead of `{ data: GetRacesResponse }` 
```suggestion
  const json = (await response.json()) as GetRacesResponse | { errors: any[] }
```

eduardopicolo-cb (changes requested):

eduardopicolo-cb (approved):

PR #1499, opened 2024-10-10
feat: creates action campaigns deeplink

### LINKED ISSUE #1497: Create multiple campaigns action dialog deeplink
We need to create some logic to open the campaigns dialog as a deeplink when the user clicks on an action on cb app that have more than one campaign

twistershark:
closes #1497 

## What changed? Why?

This PR prepares a new deeplink to be used by the CB app whenever we have more than one campaign active at the same time for a specific action.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
+++ b/src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
@@ -10,9 +10,11 @@ import { generateMetadataDetails } from '@/utils/server/metadataUtils'
 
 export const dynamic = 'error'
 
-type Props = React.PropsWithChildren<PageProps<{ action: UserActionType }>>
+type UserActionCampaignsLayoutProps = React.PropsWithChildren<PageProps<{ action: UserActionType }>>
 
-export async function generateMetadata({ params }: Props): Promise<Metadata> {
+export async function generateMetadata({
+  params,
+}: UserActionCampaignsLayoutProps): Promise<Metadata> {
   const { action } = params
   const parsedAction = action?.toUpperCase()
 
@@ -29,7 +31,10 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
   })
 }
 
-export default async function UserActionCampaignsLayout({ params, children }: Props) {
+export default async function UserActionCampaignsLayout({
+  params,
+  children,
+}: UserActionCampaignsLayoutProps) {
   const { action } = params
   const parsedAction = action?.toUpperCase()
 
--- a/src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
+++ b/src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+import { UserActionType } from '@prisma/client'
+import { Metadata } from 'next'
+import { notFound } from 'next/navigation'
+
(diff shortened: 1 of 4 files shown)


src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { notFound } from 'next/navigation'
+
+import { HomepageDialogDeeplinkLayout } from '@/components/app/homepageDialogDeeplinkLayout'
+import { USER_ACTION_CTAS_FOR_GRID_DISPLAY } from '@/components/app/userActionGridCTAs/constants/ctas'
+import { PageProps } from '@/types'
+import { generateMetadataDetails } from '@/utils/server/metadataUtils'
+
+export const dynamic = 'error'
+
+type Props = React.PropsWithChildren<PageProps<{ action: UserActionType }>>
```
Comment (lucasrmp-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx:13]: np: the type name `Props` makes it annoying to do full repo search later, let's keep the naming convention of the component name with the props suffix (`UserActionCampaignsLayoutProps`)

olavoparno (dismissed):

src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
Comment (lucasrmp-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx:1]: Didn't get this, why do we have a different page for this "all campaigns" modal? Why not have this be the default when an action has more than one active campaign? And maybe add a query param to skip the campaign selection phase when necessary

Comment (twistershark) [reply src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx:1]: So me, Matheus and Travis discussed this yesterday and we decided that we won't change our current deeplink logic.

This new deeplink **will be only used on CB app** when the current action has more than 1 campaign active at the same time. Matheus suggested that we added a '/all' after the deeplink url when we have more than 1 campaign so that we don't have to update contentful deeplink url when we have multiple campaigns. **We are going to control this addition inside the Action Card on CB app, so the logic will be controlled there.**

We don't want to change our current deeplinks for the following reasons:
- We already have social posts with deeplink for our current main campaign. When we create new campaigns, usually is for something temporary like CNN email or pizza day, while the main campaign stays active. Changing the current behavior could lead to an increased step for the user to decide which campaign he needs/wants to do when accessing the deeplink with multiple campaigns active.
- This change would be made across all actions deeplinks and could create unexpected bugs. As the only client that will be using this is the CB app, IMO it doesn't make sense for us to change our current product functionality.
- Travis didn't see much benefits on changing our current legacy deeplinks.

src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
Comment (twistershark) [reply src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx]: Updated

src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx
Comment (twistershark) [reply src/app/[locale]/(homepageDialogDeeplink)/action/[action]/all/layout.tsx:1]: We are doing this because we don't want to create new dialogs and more logic on CB app to handle multiple campaigns as we did on SWC. We are transferring the responsibility for this to SWC

lucasrmp-cb (approved):

olavoparno (approved):

PR #1522, opened 2024-10-23
chore: adjust content security policy

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-internal/issues/274

fixes 

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -27,7 +27,6 @@ const contentSecurityPolicy = {
     'https://*.googleapis.com',
     'https://*.gstatic.com',
     '*.google.com',
-    'https://www.googletagmanager.com',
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -27,7 +27,6 @@ const contentSecurityPolicy = {
     'https://*.googleapis.com',
     'https://*.gstatic.com',
     '*.google.com',
-    'https://www.googletagmanager.com',
```
twistershark, line 30: Don't we need this anymore?

twistershark (approved):

next.config.js
olavoparno: it's just reordered, it's still there

PR #1523, opened 2024-10-23
refactor: create communication journey in sms status webhook

LucasdRossi:
closes [#260](https://github.com/Stand-With-Crypto/swc-internal/issues/260)

fixes [PROD-SWC-WEB-4PP](https://stand-with-crypto.sentry.io/issues/5759605772/events/204190f7ccd34b42ac35fe761bb0803e/); [PROD-SWC-WEB-4WW](https://stand-with-crypto.sentry.io/issues/5895385971/events/0ec9070cd51d42439d18e9adc60f55ad/)

## What changed? Why?

Currently, instead of creating the communication journey and user records after sending a batch of messages, we're creating them in the status webhook. This prevents the webhook from receiving calls for messages that haven't yet been registered in our database.

## How has it been tested?

- [ ] Locally
- [x] Vercel Preview Branch
- [x] Unit test
- [ ] Functional test


LucasdRossi:
Since this will perform worse than what we have today, we’ll remove all calls to the user_communication table and pass campaign_name and journey_type via URL parameters instead. This will be done in #1553

src/app/api/public/sms/events/status/route.ts
```diff
@@ -46,48 +37,69 @@ export const POST = withRouteMiddleware(async (request: NextRequest) => {
...
+  const existingMessage = await prismaClient.userCommunication.findFirst({
+    where: {
+      messageId,
+    },
+  })
+
+  if (existingMessage) {
+    logger.info(`Found existing message with id ${messageId}`)
+    // TODO: update message status
+  } else {
```
lucasrmp-cb, line 82: qq: is it possible to leverage `prisma.upsert` instead of doing a separate query? Wanna reduce the number of db queries we do here as most as possible

lucasrmp-cb (approved):

src/utils/server/sms/communicationJourney.ts
```diff
@@ -0,0 +1,100 @@
...
+      id: true,
+      phoneNumber: true,
+    },
+  })
+
+  if (!campaignName && !journeyTypesWithSingleJourney.includes(journeyType)) {
+    throw new Error(`Please inform a campaign name for journey ${journeyType}`)
+  }
+
+  const usersWithExistingCommunicationJourney = (
```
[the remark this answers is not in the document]
LucasdRossi: This call is made to ensure that users who already have a communication journey are filtered out during creation. One approach could be to check if usersWithExistingCommunicationJourney has the same length as users, and if so, skip the next creation call, as it would mean all users already have a communication journey.

src/app/api/public/sms/events/status/route.ts
LucasdRossi: Unfortunately not, since upsert is meant for updating a single record, and we need to update multiple records that share the same `messageId`.

PR #1535, opened 2024-10-29
feat: reduces table scan on sign in database query

twistershark:
## What changed? Why?

Upon investigation, we discovered that two queries had a full table scan and these queries where executed on sign in, potentially causing a slow sign in experience for users. This PR addresses this issue by removing this table scan on the database query and using JS to filter the results afterwards.

I also identified some large queries related to authentication that were taking over 2 seconds to complete. To enhance performance during the sign-in process, I decided to break these queries into smaller ones that can be executed in parallel.

### Before the change

image

### After the change

image

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/app/api/identified-user/full-profile-info/route.ts
```diff
@@ -45,12 +44,35 @@ async function apiResponseForUserFullProfileInfo() {
     },
   })
 
+  let primaryUserEmailAddress
+  let primaryUserCryptoAddress
```
LucasdRossi, line 48: These two variables should be typed

src/app/api/identified-user/full-profile-info/route.ts
```diff
@@ -45,12 +44,35 @@ async function apiResponseForUserFullProfileInfo() {
...
+
   return {
-    user: user && {
-      ...getSensitiveDataClientUser(user),
-      address: user.address && getClientAddress(user.address),
-      userActions: user.userActions.map(record => getSensitiveDataClientUserAction({ record })),
-    },
+    user: user &&
+      primaryUserEmailAddress &&
+      primaryUserCryptoAddress && {
```
LucasdRossi, line 67: Are you sure we want to return this object if all three of these values are defined?
Shouldn’t we just check if the user is defined and let `getSensitiveDataClientUser` handle cases where `primaryUserEmailAddress` or `primaryUserCryptoAddress` are undefined?

src/app/api/identified-user/full-profile-info/route.ts
twistershark: Updated

src/app/api/identified-user/full-profile-info/route.ts
twistershark: Good suggestion! Updated!

LucasdRossi (dismissed):

twistershark:
There is still a query taking over 2 seconds to complete. This one in specific: https://github.com/Stand-With-Crypto/swc-web/blob/978e2fe6178c32d8ef737b2a5f450f54229f062a/src/app/api/identified-user/full-profile-info/route.ts#L14

Upon some investigation, I tried using pagination as an attempt to improve this query performance, by parallelising multiple smaller queries.

image


Unfortunately, it had the opposite result: 

image

lucasrmp-cb (dismissed):

twistershark:
Upon research, I found that Prisma, on version 5.8.0, [added a new relation load strategy that changes the way prisma handle relations. Instead of sending multiple queries to the database (one per table) and then joining them on the application level, the new feature uses a database-level LATERAL JOIN (PostgreSQL) and fetches all data with a single query to the database. ](https://www.prisma.io/docs/orm/prisma-client/queries/relation-queries#relation-load-strategies-preview)

[**Another important difference between these two options is that the join strategy uses JSON aggregation on the database level. That means that it creates the JSON structures returned by Prisma Client already in the database which saves computation resources on the application level.**](https://www.prisma.io/docs/orm/prisma-client/queries/relation-queries#relation-load-strategies-preview)

Results after enabling the experimental feature:
image

src/utils/server/authentication/appRouterGetAuthUser.ts
```diff
@@ -30,17 +31,20 @@ export async function appRouterGetAuthUser(): Promise<ServerAuthUser | null> {
...
           id: sessionId,
         },
       },
     },
   })
-  if (!user) {
+
+  const hasOptedIn = user?.userActions.some(action => action.actionType === UserActionType.OPT_IN)
+
+  if (!user || !hasOptedIn) {
```
eduardopicolo-cb, line 47: Do we want to return null for anonymous clients? 
What happens if an anonymous client performs an action (saves the user to the db with the sessionId), could this cause an issue somewhere, or perhaps it fixes the `'useApiResponseForUserFullProfileInfo' did not return 'primaryUserCryptoAddress' for a logged in user` error?

twistershark: I don't think there is a need to change the behavior of this function as it is already working correctly. It is used in some user actions to check if the user is authenticated and we return an error if the function returned null informing that the user is unauthenticated.

I don't have enough information about the `useApiResponseForUserFullProfileInfo` error, but this function is going to return null when the user has not joined SWC(different from anonymous users). Not sure the concern here, but I imagine that there is an action with a broken logic that is allowing users with no `primaryUserCryptoAddress` to mint NFTs, causing the error, based on the error message. 

This is a problem when we don't check if the user is logged in with Thirdweb in actions that require NFT minting. Situations like this, the user prob is logged in through userId and sessionId on query params(mobile redirect).

PR #1538, opened 2024-10-29
fix: add redis cache to voter actions count

lucasrmp-cb:
## What changed? Why?

We started seeing [this error](https://stand-with-crypto.sentry.io/issues/5712266764/events/f077b0ae98314d5798fe445249338cde/?environment=production&project=4506490717470720&referrer=previous-event&statsPeriod=30d&stream_index=1) when the first version of this went live, somehow we were trying to build the home page multiple times a second. To fix that fast without going much into why that's happening we are adding a hard redis cache layer for the metric that broke the build.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


LucasdRossi (approved):

src/inngest/functions/updateVoterActionsCountCacheCronJob.ts
```diff
@@ -0,0 +1,33 @@
...
+  'script.update-voter-actions-cache-cron-job'
+const UPDATE_VOTER_ACTIONS_COUNT_CACHE_CRON_JOB_EVENT_NAME =
+  'script/update.voter.actions.cache.cron.job'
+
+export interface UpdateVoterActionsCounterCacheCronJobSchema {
+  name: typeof UPDATE_VOTER_ACTIONS_COUNT_CACHE_CRON_JOB_EVENT_NAME
+}
+
+/**
+ * This Inngest function is a cron job responsible for backfilling the NFTs for the user actions that were skipped/missed.
```
twistershark, line 17: Is this comment related to this file?

twistershark (dismissed):

travisbloom-cb (commented):
we might want to name this cron more genericly so we can trigger multiple metrics in the same fn in the future

twistershark (approved):

travisbloom-cb (approved):

PR #1605, opened 2024-11-08
Adjust headers size & spacing

eduardopicolo-cb:
## What changed? Why?

Adjust all headings/subtitles and updates all pages to use `PageTitle` and `PageSubtitle` components

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/ui/pageTitleText/index.tsx
```diff
@@ -28,10 +31,34 @@ export interface PageTitleProps
...
+      return 'xs'
+    default:
+      return DEFAULT_PAGE_TITLE_SIZE
+  }
+}
+
 export const PageTitle = React.forwardRef<HTMLHeadingElement, PageTitleProps>(
   ({ className, children, as: Comp = 'h1', size, withoutBalancer, ...props }, ref) => {
+    const computedSize = size || getSizeFromComp(Comp)
+    console.log('computedSize:', computedSize)
```
twistershark, line 55: we don't need this

src/components/app/footer/index.tsx
```diff
@@ -22,7 +22,7 @@ export async function Footer({ locale }: { locale: SupportedLocale }) {
       <div className="container">
         <div className="flex flex-col gap-9 lg:flex-row lg:justify-between">
           <div className="max-w-2xl space-y-8">
-            <PageTitle as="h6" className="text-left">
+            <PageTitle as="h6" className="text-left" size={DEFAULT_PAGE_TITLE_SIZE}>
```
twistershark, line 25: As the `DEFAULT_PAGE_TITLE_SIZE` is the default prop value for size, do we still need to pass it here?

src/components/app/pageAbout/index.tsx
```diff
@@ -22,7 +22,9 @@ export function PageAbout({ title, description }: { title: string; description:
           ].map(({ value, label }) => (
             <Card key={label}>
               <Card.Heading>
-                <PageTitle as="p">{value}</PageTitle>
+                <PageTitle as="p" size={DEFAULT_PAGE_TITLE_SIZE}>
```
twistershark, line 25: As the `DEFAULT_PAGE_TITLE_SIZE` is the default prop value for size, do we still need to pass it here?

src/components/app/footer/index.tsx
eduardopicolo-cb: Yes, needed to overwrite the `h6` size and have the same size as a `PageTitle` without being an `h1` tag

src/components/app/pageAbout/index.tsx
eduardopicolo-cb: Yes, needed to overwrite the `p` size

twistershark (approved):

PR #1620, opened 2024-11-22
fix: politician page scroll

### LINKED ISSUE #814: Fix scroll position at the bottom when accessing a politician page on mobile
Sometimes when you access a politician page, scroll down, go back and access another one, the scroll is set at the bottom of the page. This usually happens on mobile and when the first politician page had a lot of content while the other one doesn't.

https://github.com/Stand-With-Crypto/swc-web/assets/165944245/4aac4ca9-c8e7-447c-9a80-e8d02132838d

Possible solution: https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#disabling-scroll-restoration

twistershark:
closes #814 

## What changed? Why?

This PR addresses a long-standing issue where, on mobile devices in production, if you visit a politician's page, scroll down, then navigate back and to a different politician's page, the scroll position from the previous page is unexpectedly carried over to the new page. This requires the user to manually scroll back to the top.

I've tested various approaches to fix this, including [enabling and disabling scroll behavior on router.push](https://nextjs.org/docs/app/api-reference/functions/use-router#disabling-scroll-to-top), but this didn’t resolve the issue. I also tried adding an ID to the politician photo section and appending it to the router's pathname, but that approach didn't work either. Given that all politician page components are server-side rendered and useEffect can’t be used to handle scrolling, I created a new component that ensures the page scrolls to the top upon render. This solution can also be reused in other server-side components.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

## Before the fix

https://github.com/user-attachments/assets/d6b55c44-1638-4947-9447-a03e6998586c

## After the fix

https://github.com/user-attachments/assets/125f142c-4637-43ad-8ee2-8cfb5bba5ec8

### DIFF
--- a/cypress/e2e/09-page-politician-details.cy.ts
+++ b/cypress/e2e/09-page-politician-details.cy.ts
@@ -6,9 +6,7 @@ describe('page - politician details', () => {
 
     cy.get('h2').contains('Candidate questionnaire')
 
-    cy.wait(3000)
-
-    cy.get('[data-testid="questionnaire-trigger"]').click()
+    cy.get('[data-testid="questionnaire-trigger"]').click({ force: true })
 
     cy.get('div')
       .should(
--- a/cypress/e2e/09-page-politician-details.cy.ts
+++ b/cypress/e2e/09-page-politician-details.cy.ts
@@ -6,6 +6,8 @@ describe('page - politician details', () => {
 
     cy.get('h2').contains('Candidate questionnaire')
 
+    cy.wait(3000)
+
     cy.get('[data-testid="questionnaire-trigger"]').click()
 
     cy.get('div')
--- a/cypress/e2e/09-page-politician-details.cy.ts
+++ b/cypress/e2e/09-page-politician-details.cy.ts
@@ -6,7 +6,7 @@ describe('page - politician details', () => {
 
     cy.get('h2').contains('Candidate questionnaire')
 
-    cy.contains('Responses (8)').click()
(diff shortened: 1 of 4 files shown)


cypress/e2e/09-page-politician-details.cy.ts
```diff
@@ -6,7 +6,9 @@ describe('page - politician details', () => {
 
     cy.get('h2').contains('Candidate questionnaire')
 
-    cy.contains('Responses (8)').click()
+    cy.wait(3000)
```
lucasrmp-cb, line 9: Do we need this wait? Doesn't the timeout on the `cy.get` call below cover this element not being rendered?

twistershark: For some reason, only on the E2E tests, the click on the accordion trigger was not opening the questionnaire. 

Spent hours investigating this and didn't find the reason. Then I thought that maybe the click could be happening at the same time as the scroll, causing a lost of focus or something like that. That's why I added this wait to try to test if something like this was really happening and it actually solved the failing E2E test

eduardopicolo-cb: Have you tried `.click({ force: true })`? 
It bypasses cypress` visibility checks, which can block clicks when the window is scrolling

eduardopicolo-cb (approved):

lucasrmp-cb (approved):

Issue #815, opened 2024-05-03
Prevent the politician row from navigating when clicking while holding CMD or CTRL


ghost:
When you try to open a politician's page in a new tab using CMD or CTRL, even though the new tab is opened, the table row is also capturing this click and navigating the current tab to the politician's page.

https://github.com/Stand-With-Crypto/swc-web/assets/165944245/01f0a4a8-4bd2-4436-8394-02d08fd064bc

travisbloom-cb:
We'll want to ensure this also continues to support the entire row being clickable. (We may want to see if we can somehow wrap the content in a a tag instead of relying on onClick events)

olavoparno:
linked issues:

https://github.com/Stand-With-Crypto/swc-web/issues/450
https://github.com/Stand-With-Crypto/swc-web/issues/453

PR #1623, opened 2024-11-26
style: adjust politician table and other stuff

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/815

fixes 

## What changed? Why?

- politician filters now on query parameters
- click on politicians while holding ctrl or cmd do not navigate
- menu zindex adjusted

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -82,11 +100,47 @@ export function DataTable<TData extends Person = Person>({
...
+
   const router = useRouter()
+  const pathname = usePathname()
+
+  useEffect(() => {
+    if (!pathname || !searchParams) {
+      return
+    }
+
+    const currentParams = new URLSearchParams(searchParams?.toString() ?? '')
```
LucasdRossi, line 128: Do you really need to instantiate URLSearchParams again here? Can't you use `params` from line 105?

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -116,9 +170,65 @@ export function DataTable<TData extends Person = Person>({
...
+    }
+
+    window.addEventListener('resize', handleResize)
+    window.addEventListener('scroll', handleResize)
+
+    return () => {
+      window.removeEventListener('resize', handleResize)
+      window.removeEventListener('scroll', handleResize)
+    }
+  }, [])
```
LucasdRossi, line 227: I would move this to a separate hook: `useResize`

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -141,59 +252,92 @@ export function DataTable<TData extends Person = Person>({
...
+                            ? null
+                            : flexRender(header.column.columnDef.header, header.getContext())}
+                        </TableHead>
+                      )
+                    })}
+                  </TableRow>
+                ))}
+              </TableHeader>
+              <TableBody>
+                {table.getRowModel().rows?.length ? (
```
LucasdRossi, line 303: Move `table.getRowModel()` to a variable outside the rendering logic

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -141,59 +252,92 @@ export function DataTable<TData extends Person = Person>({
...
+                      onClick={event => {
+                        const politicianUrl = getIntlUrls(locale).politicianDetails(
+                          row.original.slug,
+                        )
+                        if (event.ctrlKey || event.metaKey) {
+                          event.preventDefault()
+                          return window.open(politicianUrl, '_blank')
+                        }
+                        return router.push(politicianUrl)
+                      }}
```
LucasdRossi, line 318: Move this to a function outside the render logic

lucasrmp-cb:
This closes more than one issue if I'm not mistaken, right? Let's add the `closes` keyword for all of them

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
lucasrmp-cb, line 227: For the resize logic we have this in the `useResizeObserver` hook, all you need yo do is pass it with the document ref and it should give you width and height of the bounding client rect.

And also, why would the width change on scroll?

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -82,11 +100,47 @@ export function DataTable<TData extends Person = Person>({
...
+
+      return currentParams.set(key, stringify ? JSON.stringify(value) : value)
+    }
+
+    updateParams(columnFiltersKey, columnFilters, true)
+    updateParams(sortingKey, sorting, true)
+    updateParams(globalFilterKey, globalFilter, false)
+
+    router.replace(`${pathname}?${currentParams.toString()}`, { scroll: false })
+  }, [columnFilters, globalFilter, pathname, router, searchParams, sorting])
```
lucasrmp-cb, line 143:
```
https://swc-web-git-refactor-style-n-other-stuff-coinbase-vercel.vercel.app/politicians?filters=%5B%7B%22id%22%3A%22swcStanceScore%22%2C%22value%22%3A%22All%22%7D%2C%7B%22id%22%3A%22primaryRole%22%2C%22value%22%3A%22All%22%7D%2C%7B%22id%22%3A%22politicalAffiliationCategory%22%2C%22value%22%3A%22All%22%7D%2C%7B%22id%22%3A%22state%22%2C%22value%22%3A%22All%22%7D%5D&sorting=%5B%5D&globalFilter=
```

This is the url we're creating when opening the page when the only thing we're trying to do is persist 5 states in search params. I understand that the way the data table works is different than just states but why can't we create a controled state and add a persistence layer on top of it to only have the 5 query params instead of a bunch of unreadable text on the url?

The way I'd think about this issue would be to first make sure we have a controlled state of the filters and sorting (this you already have here), and then create a `useSearchParamsState` that would be the persistence layer between the query params and the state (explanation below)

The hook would receive a key an a default value as arguments, have the exact same return type as `useState` and work like this:
- It would have a setState in which it would check if the search params with that `key` is set and if so use it as default value and fallback to the one provided by the call (you can do this by passing a function as an argument to `useState`)
- It would wrap the `setState` function where it will set both the state from the `useState` call and also the `key` in the search params (without needing a reload since the filter value would be up to date)

We can pair on this if you like

lucasrmp-cb (changes requested):

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: Even though this handling two different events, I'm abstracting this to a separate hook. Thanks man

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: Thanks man

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: It has to be recalculated on scroll to get the correct sum of header widths because it loses its style after changing the position to `fixed` and that's why I can't use the useResizeObserver. The header is virtually recreated because of this

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: thanks for the pairing man

src/hooks/useQueryParamState.ts
```diff
@@ -1,42 +1,60 @@
...
+    }
+
+    return defaultValue
+  })
+
+  const setState: Dispatch<SetStateAction<Value>> = useCallback(
+    newValue => {
+      const valueToSet = typeof newValue === 'function' ? newValue(defaultValue) : newValue
+
+      setInternalState(valueToSet)
```
lucasrmp-cb, line 38: This is not necessary if you are only setting the internal state, just return `setInternalState` from the hook and it will work the same

src/components/app/dtsiClientPersonDataTable/useTableFilters.ts
```diff
@@ -0,0 +1,126 @@
...
+import { PERSON_TABLE_COLUMNS_IDS } from '@/components/app/dtsiClientPersonDataTable/columns'
+import {
+  PARTY_OPTIONS,
+  ROLE_OPTIONS,
+  StanceOnCryptoOptions,
+} from '@/components/app/dtsiClientPersonDataTable/globalFiltersUtils'
+import { useSearchParamState } from '@/hooks/useQueryParamState'
+
+export function useGlobalFilter(): [string, Dispatch<SetStateAction<string | undefined | null>>] {
+  const [searchValue, setSearchValue] = useSearchParamState('global')
```
lucasrmp-cb, line 15: Can we change the key here to something like `search`/`q`/`query`? It's a common pattern

lucasrmp-cb, line 15: We should have default values on this as an empty string, no?

src/components/app/dtsiClientPersonDataTable/useTableFilters.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  return [searchValue ?? '', setSearchValue]
+}
+
+export function useSortingFilter(): [SortingState, Dispatch<SetStateAction<SortingState>>] {
+  const [sortValue, setSortValue] = useSearchParamState('sorting')
```
lucasrmp-cb, line 21: Same here

lucasrmp-cb (changes requested):

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -116,9 +123,71 @@ export function DataTable<TData extends Person = Person>({
...
+
+  const filterContainerBoundingRect = filterContainerRef.current?.getBoundingClientRect()
+  const currentFilterContainerYPx = filterContainerBoundingRect
+    ? filterContainerBoundingRect.top + filterContainerBoundingRect.height
+    : 0
+
+  const shouldShowFixedHeader = !isSearchContainerVisible && !isPaginationVisible
+
+  const headerRefs = useRef<HTMLTableCellElement[]>([])
+  const [headerWidths, setHeaderWidths] = useState<number[]>([])
```
LucasdRossi, line 155: It's a good idea to keep all your useState calls together at the top of the function. I also like to group hooks by type, like putting all the useEffect and useMemo calls together. That way, it's easy to see all the states, memoized variables, and effects in one spot.

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: no longer on the PR

src/hooks/useQueryParamState.ts
olavoparno: thanks

src/components/app/dtsiClientPersonDataTable/useTableFilters.ts
olavoparno: thanks

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: thanks

src/hooks/useQueryParamState.ts
lucasrmp-cb, line 1: Let's change either the name of the file or the hook so they are the same

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

Issue #1529, opened 2024-10-25
Upgrade to Next 15


lucasrmp-cb:
Next 15 launched recently and the objective here is to upgrade the package and it's dependencies.

https://nextjs.org/docs/app/building-your-application/upgrading/version-15

Theoretically all we need to update is next, react, react-dom and eslint (see the upgrade guide above) but with the amount of breaking changes this major introduces we will need to research compatibility and upgrade some dependencies if necessary (my bigger concern here is our i18n logic and if it works with the new async params from next 15, we might need a completely different solution or wait for compatibility)

Most changes required should be covered by the [upgrade guide](https://nextjs.org/docs/app/building-your-application/upgrading/version-15) and it also have a codemod cli that should make most of the bulk changes for us.

Besides doing extensive QA on this change we also want to measure the impact this has on development server performance that has been an issue lately.

olavoparno:
Other tasks as a reference: 

#1350

olavoparno:
bugbash https://docs.google.com/document/d/1u6gKY95LPa_wya1kzNhiIPElf3HOlbBn4lqIA0PVRds/edit?tab=t.0
tdd follow up https://docs.google.com/document/d/1FC1yYKEoKAtOgYyo48cuc3PwGdq59JoganMOMqAvo54/edit?tab=t.0

PR #1628, opened 2024-11-29
Chore/upgrade next 15

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1529

fixes 

## What changed? Why?

**Upgraded our project to Next 15 new API.**

Please, run `rm -rf .next node_modules package-lock.json && npm i` before testing it locally

**TLDR**: We can now leverage everything from the latest React and Next API changes and with this we have **faster** DEV coding (**not bugging anymore!**), faster build time and **we may finally resolve the `user session cookie not set` [errors](https://stand-with-crypto.sentry.io/issues/?project=4506490717470720&query=user%20session%20cookie%20not%20set&referrer=issue-list&statsPeriod=30d)** since cookies and headers are now async functions. [Link to issue](https://github.com/Stand-With-Crypto/swc-web/issues/1490).

Following the upgrade [guide](https://nextjs.org/docs/app/building-your-application/upgrading/version-15) and a couple of links listed below, here's some important points on it:

**Notes** (according to new API):

- The revalidate value needs to be _**statically analyzable.**_ For example revalidate = 600 is valid, but revalidate = 60 * 10 is not.
- The revalidate value is not available when using runtime = 'edge'.
- If the **dynamic = 'error' and dynamic = 'force-static'** are used, it'll change the default of **dynamicParams to false.**
- We can now make use of the new **empowered** [**instrumentation**](https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation#onrequesterror-optional) file already integrated with [Sentry](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#errors-from-nested-react-server-components).
- We have a warning being shown on building which is not a problem itself but we should look into it. 
image

- [**Sentry does not support turbopack yet**](https://github.com/getsentry/sentry-javascript/issues/8105) so no local Sentry integration (currently not a problem for us since we disable it locally).

**Attention**:

> The use of dynamic and dynamicParams as we had is no longer available since it does not even allow building. Everything related to route/page options must be as the doc says "statically analyzable", so we can't use variables from ENV and asserting values conditionally, so we may have to depend on the very functions to OPT out from building pages or not, like generateStaticParams. **Must double check its behavior while building**.

**Next steps**:

- Thoroughly test the app also trying to leverage our newly built synthetic tests, test out static/dynamic pages, actions, deeplinks, donation, sign in/signup, check on sentry for errors, vercel logs, try triggering them manually as well, etc.
- Upgrade other dependencies.
- Remove package.json's **overrides** as libs upgrade to React 19.

**Links**:

- [codemods](https://nextjs.org/docs/app/building-your-application/upgrading/codemods)
- [route options](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#options)
- [react 19 ts](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#typescript-changes)
- [motion for react](https://motion.dev/docs/react-quick-start#react-19-next-15)

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/dtsiPersonHeroCard/dtsiPersonHeroCardSection.tsx
```diff
@@ -14,7 +14,7 @@ export interface DTSIPersonHeroCardSectionProps {
   title: React.ReactNode
   titleProps?: PageTitleProps
   subtitle?: React.ReactNode
-  cta?: ReactElement
+  cta?: ReactElement<any>
```
mhamendes, line 17: We should probablty type this as `React.ReactNode` as well. If that is not possible, we should avoid using `any` and use `unknown` instead.

src/components/app/pageAdvocatesHeatmap/constants.ts
```diff
@@ -1,3 +1,4 @@
+import type { JSX } from 'react'
```
mhamendes, line 1: Probably should be typed as `React.ReactNode` as well.

src/components/app/pageEvents/eventsPageDialogDeeplinkLayout.tsx
```diff
@@ -18,16 +18,17 @@ import { getIntlUrls } from '@/utils/shared/urls'
...
+  pageParams: Awaited<PageProps['params']>
   hideModal?: boolean
-  children: ReactElement
+  children: ReactElement<any>
 }
 
 export async function EventsPageDialogDeeplinkLayout({
   children,
   pageParams,
 }: EventsPageDialogDeeplinkLayout) {
```
mhamendes, line 29: Children should definitely should be type as `React.ReactNode`, but better yet, `EventsPageDialogDeeplinkLayout` should be wrapped by `React.PropsWithChildren`, which types children correctly.

Something like this:
```js
interface EventsPageDialogDeeplinkLayout {
  pageParams: Awaited<PageProps['params']>
  hideModal?: boolean
}

export async function EventsPageDialogDeeplinkLayout({
  children,
  pageParams,
}: React.PropsWithChildren<EventsPageDialogDeeplinkLayout>) {
```

src/components/app/dtsiPersonHeroCard/dtsiPersonHeroCardSection.tsx
olavoparno: This was prob added from react upgrade's codemods, we should add a rule on our lint to avoid this

src/components/app/pageEvents/eventsPageDialogDeeplinkLayout.tsx
olavoparno: Here we prob are going with JSX.Element because of cloneElement's typings

src/components/app/pageDonate/donateButton.tsx
```diff
@@ -22,14 +22,15 @@ export function DonateButton() {
         },
       },
       () =>
-        actionCreateCoinbaseCommerceCharge().then(res => {
-          if (res && 'hostedUrl' in res) {
-            const windowReference = openWindow(res.hostedUrl, '_blank', '')
+        actionCreateCoinbaseCommerceCharge().then(async res => {
+          const commerceChargeResponse = await res
```
twistershark, line 26: np: Naming the `res` variable to `getCoinbaseCommerceCharge` might make it easier to understand, especially since we need to await it now

src/utils/shared/phoneNumber.ts
```diff
@@ -22,15 +24,26 @@ export function normalizePhoneNumber(passed: string) {
 export function formatPhoneNumber(phoneNumber: string) {
   if (!phoneNumber) return ''
 
-  const parsedPhoneNumber = parsePhoneNumber(phoneNumber, DEFAULT_COUNTRY_CODE)
+  // https://github.com/catamphetamine/libphonenumber-js/issues/468#issue-2504182999
```
twistershark, line 27: np: Could you add a brief description of why you linked to this issue here? I tried opening the issue, but without knowing what I should be looking for, it’s a bit hard to understand what the issue is.

Thanks!

src/components/app/pageDonate/donateButton.tsx
olavoparno: opa, thanks man. you're right but we could work on that since it is a sideeffect from the API changing. to be honest, i'm not happy with this new way of having to await for the action response :( I couldn't find a better way to handle cookies and headers on the actions middleware

twistershark (approved):

src/utils/shared/phoneNumber.ts
olavoparno: thanks man, i've added extra info!

twistershark (dismissed):

src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx
```diff
@@ -2,13 +2,13 @@ import { HomepageDialogDeeplinkLayout } from '@/components/app/homepageDialogDee
 import { UserActionFormCallCongresspersonDeeplinkWrapper } from '@/components/app/userActionFormCallCongressperson/homepageDialogDeeplinkWrapper'
 import { dialogContentPaddingStyles } from '@/components/ui/dialog/styles'
 import { PageProps } from '@/types'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 import { cn } from '@/utils/web/cn'
 
-export const revalidate = SECONDS_DURATION.HOUR
+export const revalidate = 3600 // 1 hour
```
Comment (lucasrmp-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx:7]: Use the shared util here

src/app/[locale]/(homepageDialogDeeplink)/action/email-debate/page.tsx
```diff
@@ -1,12 +1,12 @@
 import { HomepageDialogDeeplinkLayout } from '@/components/app/homepageDialogDeeplinkLayout'
 import { UserActionFormEmailDebateDeeplinkWrapper } from '@/components/app/userActionFormEmailDebate/homepageDialogDeeplinkWrapper'
 import { PageProps } from '@/types'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 
-export const revalidate = SECONDS_DURATION.HOUR
+export const revalidate = 3600 // 1 hour
```
Comment (lucasrmp-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/email-debate/page.tsx:5]: +1

src/app/[locale]/(homepageDialogDeeplink)/action/live-event/[slug]/page.tsx
```diff
@@ -7,18 +7,18 @@ import { UserActionFormLiveEventDeeplinkWrapper } from '@/components/app/userAct
 import { dialogContentPaddingStyles } from '@/components/ui/dialog/styles'
 import { PageProps } from '@/types'
 import { generateMetadataDetails } from '@/utils/server/metadataUtils'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 import { UserActionLiveEventCampaignName } from '@/utils/shared/userActionCampaigns'
 
-export const revalidate = SECONDS_DURATION['30_SECONDS']
+export const revalidate = 30 // 30 seconds
```
Comment (lucasrmp-cb) [inline src/app/[locale]/(homepageDialogDeeplink)/action/live-event/[slug]/page.tsx:12]: +1

docs/Add a new user action.md
```diff
@@ -175,7 +175,7 @@ export function UserActionFormActionNameSkeleton() {
 ```javascript
 /// page.tsx
 
-export const revalidate = SECONDS_DURATION.HOUR
+export const revalidate = 900 // 15 minutes
```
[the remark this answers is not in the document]
olavoparno: We can't now, it's a Next constraint. Ive added to the PR description; Next says with revalidate, everything has to be _statically analyzable_

package.json
```diff
@@ -6,7 +6,7 @@
     "ts": "TZ=UTC dotenv -- tsx --tsconfig ./tsconfig.bin.json",
     "ts:ci": "TZ=UTC tsx --tsconfig ./tsconfig.bin.json",
     "dev": "next dev",
-    "dev:turbo": "next dev --turbo",
```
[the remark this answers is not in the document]
olavoparno: We can leave "dev" pointing towards turbopack by default but since we have Sentry with webpack config Next will fallback to webpack too if not explicit on the command :( also, Sentry does not support turbopack yet, so prob sticking with the --turbo on the command

src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx
Comment (olavoparno) [reply src/app/[locale]/(homepageDialogDeeplink)/action/call/page.tsx:7]: Same for the _statically analyzable_

src/app/[locale]/(homepageDialogDeeplink)/action/email-debate/page.tsx
Comment (olavoparno) [reply src/app/[locale]/(homepageDialogDeeplink)/action/email-debate/page.tsx:5]: Same for the _statically analyzable_

src/app/[locale]/(homepageDialogDeeplink)/action/email/page.tsx
```diff
@@ -1,12 +1,12 @@
 import { HomepageDialogDeeplinkLayout } from '@/components/app/homepageDialogDeeplinkLayout'
 import { UserActionFormEmailCongresspersonDeeplinkWrapper } from '@/components/app/userActionFormEmailCongressperson/homepageDialogDeeplinkWrapper'
 import { PageProps } from '@/types'
-import { SECONDS_DURATION } from '@/utils/shared/seconds'
 
-export const revalidate = SECONDS_DURATION.HOUR
+export const revalidate = 3600 // 1 hour
```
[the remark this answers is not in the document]
Comment (olavoparno) [reply src/app/[locale]/(homepageDialogDeeplink)/action/email/page.tsx:5]: Same for the _statically analyzable_

src/app/[locale]/(homepageDialogDeeplink)/action/live-event/[slug]/page.tsx
Comment (olavoparno) [reply src/app/[locale]/(homepageDialogDeeplink)/action/live-event/[slug]/page.tsx:12]: Same for the _statically analyzable_

lucasrmp-cb (approved):

twistershark (approved):

PR #1637, opened 2024-12-05
chore: try checking db schema before building

olavoparno:
closes 

fixes 

## What changed? Why?

checks for db schemas differences and fails on Vercel build if so
only runs on production

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


package.json
```diff
@@ -23,6 +23,7 @@
     "intl:extract": "formatjs extract 'src/**/*.{ts,tsx}' '!src/**/*.d.ts' --out-file src/intl/generated/extract.json",
     "intl:extract-compile": "formatjs extract 'src/**/*.ts*' '!src/**/*.d.ts' --out-file temp.json --flatten --id-interpolation-pattern '[sha512:contenthash:base64:6]' && formatjs compile 'temp.json' --out-file src/intl/generated/compile/en-US.json && rm temp.json",
     "db:seed": "prisma db push --force-reset && prisma db seed",
+    "db:check": "if [ \"$NEXT_PUBLIC_ENVIRONMENT\" = \"production\" ]; then prisma migrate diff --from-schema-datasource=./prisma/schema.prisma --to-schema-datamodel=./prisma/schema.prisma --exit-code; else echo \"Skipping db:check because NEXT_PUBLIC_ENVIRONMENT is not production.\"; fi",
```
lucasrmp-cb, line 26: If possible we should do this check inside the `vercel_deploy.sh` script and not here, we might want to run the check locally for some reason

lucasrmp-cb (changes requested):

package.json
olavoparno: agreed man, can you check again?

PR #1638, opened 2024-12-05
feat: improves datadog cleanup function and decreases cron job schedule

### LINKED ISSUE #1635: Decrease synthetic tests cleanup cronjob period
We aim to reduce the frequency of the synthetic tests cleanup cron job. Additionally, we want to ensure that it only deletes emails that were created more than 30 minutes ago. This measure will help prevent potential bugs caused by the removal of emails while synthetic tests are still running.

twistershark:
closes #1635 

## What changed? Why?

This PR changes the cron job schedule from daily at 3 AM to hourly execution. Additionally, the synthetic tests are now run every 10 minutes. The code has also been refactored to improve maintainability.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/inngest/functions/datadog/cleanup/index.ts
+++ b/src/inngest/functions/datadog/cleanup/index.ts
@@ -1,9 +1,14 @@
-import { UserActionType } from '@prisma/client'
+import { User, UserAction, UserActionType, UserEmailAddress } from '@prisma/client'
 import { uniqBy } from 'lodash-es'
 
 import { inngest } from '@/inngest/inngest'
 import { onScriptFailure } from '@/inngest/onScriptFailure'
 import { prismaClient } from '@/utils/server/prismaClient'
+
+type UserWithRelations = User & {
+  userActions: UserAction[]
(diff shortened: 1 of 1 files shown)


src/inngest/functions/datadog/cleanup/index.ts
```diff
@@ -32,65 +37,67 @@ export const cleanupDatadogSyntheticTestsWithInngest = inngest.createFunction(
...
+
+async function updateUser(user: User) {
+  return prismaClient.user.update({
+    where: { id: user.id },
+    data: {
+      primaryUserEmailAddressId: undefined,
+      primaryUserEmailAddress: undefined,
+    },
+  })
+}
```
lucasrmp-cb, line 77: Isn't this open to the same timing issues we discussed previously? Even though we're not removing the emails we're unassigning it from the user which has the same effect in practice

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

PR #1640, opened 2024-12-09
fix: check if has hydrated before rendering logged in content

LucasdRossi:
## What changed? Why?

This PR introduces validations to ensure the page is fully hydrated before rendering content for logged-in users. This will address the following errors:
* https://stand-with-crypto.sentry.io/issues/5876226755/events/3bf117c9eb3c404e944247b6f600d145/
* https://stand-with-crypto.sentry.io/issues/5876226755/events/45d9fe878e3d488e975fbe16b36e0313/

## How has it been tested?

- [ ] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionFormNFTMint/sections/checkout/index.tsx
+++ b/src/components/app/userActionFormNFTMint/sections/checkout/index.tsx
@@ -6,6 +6,7 @@ import { getContract, PreparedTransaction } from 'thirdweb'
 import { base } from 'thirdweb/chains'
 import { TransactionButton, useActiveWalletConnectionStatus } from 'thirdweb/react'
 
+import { MaybeAuthenticatedContent } from '@/components/app/authentication/maybeAuthenticatedContent'
 import {
   NFTDisplay,
   NFTDisplaySkeleton,
@@ -30,7 +31,6 @@ import { PageSubTitle } from '@/components/ui/pageSubTitle'
 import { PageTitle } from '@/components/ui/pageTitleText'
 import { Skeleton } from '@/components/ui/skeleton'
 import { UseSectionsReturn } from '@/hooks/useSections'
-import { useSession } from '@/hooks/useSession'
 import { SupportedCryptoCurrencyCodes } from '@/utils/shared/currency'
 import { NFTSlug } from '@/utils/shared/nft'
 import { thirdwebClient } from '@/utils/shared/thirdwebClient'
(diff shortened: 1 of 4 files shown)


src/hooks/useHasHydrated.ts
```diff
@@ -2,9 +2,9 @@ import React from 'react'
 
 // fix for https://nextjs.org/docs/messages/react-hydration-error
 export const useHasHydrated = () => {
-  const [hasMounted, setHasMounted] = React.useState(false)
+  const hasMounted = React.useRef(false)
```
twistershark, line 5: Good change :D

twistershark (dismissed):

src/hooks/useHasHydrated.ts
olavoparno: just a question, have you guys tested it out thoroughly? Because on Nextjs link they use the useState solution https://nextjs.org/docs/messages/react-hydration-error#solution-1-using-useeffect-to-run-on-the-client-only

olavoparno: try running it locally 
image

src/components/app/userActionFormNFTMint/sections/checkout/index.tsx
```diff
@@ -87,145 +86,152 @@ export function UserActionFormNFTMintCheckout({
   const maybeOverriddenCheckoutError = debug ? null : checkoutError
   const connectionStatus = useActiveWalletConnectionStatus()
 
-  if (!session.isLoggedInThirdweb) {
-    return <UserActionFormNFTMintCheckoutSkeleton />
-  }
```
lucasrmp-cb, line 92: Why are we removing this? This is not the same check `MaybeAuthenticatedContent` does, this only checks for thirdweb authentication

src/hooks/useHasHydrated.ts
lucasrmp-cb: Yes, even though ref seems the logical solution it can cause hydration problems due to it changing UI without triggering a tree re-render. Unless this implementation was causing issues I'd leave as it was, if we need to change to this we need to test a big sample of use cases

lucasrmp-cb (changes requested):

olavoparno (approved):

PR #1667, opened 2024-12-17
Fix double profile modal

### LINKED ISSUE #1665: Double trigger on "Finish your profile" on the home page
https://coinbase.enterprise.slack.com/files/U068E3U6E4W/F085GPYMYH0/screenshare_-_2024-12-17_10_14_38_am.webm

eduardopicolo-cb:
closes #1665 

fixes 

## What changed? Why?

The UpateProfileForm shows twice due to two instances being mounted on the page (1 mobile & 1 desktop). The solution is to use javascript to conditionally mount the components.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/pageUserProfile/index.tsx
+++ b/src/components/app/pageUserProfile/index.tsx
@@ -82,11 +82,7 @@ export function PageUserProfile({ params, user }: PageUserProfile) {
             </div>
           </div>
 
-          {!isMobile && (
-            <div className="hidden lg:flex">
-              <ProfileAndNFTButtons user={user} />
-            </div>
-          )}
+          {!isMobile && <ProfileAndNFTButtons user={user} />}
         </div>
         <div className="grid grid-cols-3 rounded-3xl bg-secondary p-3 text-center sm:p-6">
           {[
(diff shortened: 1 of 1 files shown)


src/components/app/pageUserProfile/index.tsx
```diff
@@ -76,9 +81,12 @@ export function PageUserProfile({ params, user }: PageUserProfile) {
               </div>
             </div>
           </div>
-          <div className="hidden lg:flex">
-            <ProfileAndNFTButtons user={user} />
-          </div>
+
+          {!isMobile && (
+            <div className="hidden lg:flex">
```
twistershark, line 86: If we are adding this comparison, I think that we can remove the div as it was trying to do the same

src/components/app/pageUserProfile/index.tsx
```diff
@@ -122,9 +130,11 @@ export function PageUserProfile({ params, user }: PageUserProfile) {
         </div>
       </section>
 
-      <div className="w-full lg:hidden">
-        <ProfileAndNFTButtons user={user} />
-      </div>
+      {isMobile && (
+        <div className="w-full lg:hidden">
```
twistershark, line 134: If we are adding this comparison, I think that we can remove the div as it was trying to do the same

twistershark (approved):

olavoparno (approved):

PR #1682, opened 2024-12-19
Fix signin button

### LINKED ISSUE #1681: Signin button stuck on wrong state

eduardopicolo-cb:
closes #1681 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/navbar/navbarLoggedInButton/index.tsx
+++ b/src/components/app/navbar/navbarLoggedInButton/index.tsx
@@ -3,6 +3,7 @@
 import { useCallback, useMemo, useState } from 'react'
 import { useEvent } from 'react-use'
 
+import { LoginDialogWrapper } from '@/components/app/authentication/loginDialogWrapper'
 import { Button } from '@/components/ui/button'
 import { useResponsivePopover } from '@/components/ui/responsivePopover'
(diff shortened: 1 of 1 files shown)


twistershark (approved):

src/components/app/navbar/navbarLoggedInButton/index.tsx
```diff
@@ -57,7 +58,15 @@ export function NavbarLoggedInButton({ onOpenChange }: { onOpenChange: (open: bo
             <div className="max-w-[150px] truncate">{displayName}</div>
           </Button>
         ) : (
-          <Button data-testid="login-button">Sign In</Button>
+          <LoginDialogWrapper authenticatedContent={null}>
```
olavoparno, line 61: if we're using this wrapper why not reduce the conditionals by adding displayName in `authenticatedContent`?

eduardopicolo-cb: the wrapper is a escape hatch for that edge case, because the navbarLoggedInButton is already inside of an `authenticatedContent`.  This workaround is to allow an user to reconnect

PR #1683, opened 2024-12-19
feat: internationalization POC

### LINKED ISSUE #1656: POC on international i18n library approach
We want to make a POC checking if the i18n library is able to recognize static folders as locale, so the idea here is to have a `en-US` and one more language as static folders and not have the `[locale]` dynamic route as a entrypoint and check if that works and what changes we need to do to achieve that

twistershark:
closes #1656 

## What changed? Why?

This PR creates a proof of concept (POC) to leverage the user's locale to render different pages or components based on the user's locale. List of changes:

- Added UK locale
- Created server and client `LocalizedComponent` to render specific components based on the current user locale
- Created a mock UK events page to showcase how it would work in a real scenario
- Added a new locale property to the `User` and `UserAction` models to help filter advocates and actions based on locale.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


prisma/schema.prisma
```diff
@@ -130,6 +130,7 @@ model User {
   acquisitionSource          String                    @map("acquisition_source")
   acquisitionMedium          String                    @map("acquisition_medium")
   acquisitionCampaign        String                    @map("acquisition_campaign")
+  locale                     String?                   @map("locale")
```
twistershark, line 133: This will be optional at first until we backfill all current users to en-US. Then, we will make it required so that new users must set a specific locale.

prisma/schema.prisma
```diff
@@ -294,6 +295,7 @@ model UserAction {
   nftMintId           String?            @unique @map("nft_mint_id")
   nftMint             NFTMint?           @relation(fields: [nftMintId], references: [id])
   actionType          UserActionType     @map("action_type")
+  locale              String?            @map("locale")
```
twistershark, line 298: This will be optional at first until we backfill all current actions to en-US. Then, we will make it required so that new actions must set a specific locale.

src/app/[locale]/events/page.tsx
```diff
@@ -19,12 +22,18 @@ export const metadata: Metadata = {
...
   if (!events || !events?.length) {
     notFound()
   }
 
-  return <EventsPage events={events} />
+  if (locale === SupportedLocale.EN_US) {
+    return <EventsPage events={events} />
+  }
+
+  return <ExampleUkPageEvents />
```
Comment (twistershark) [inline src/app/[locale]/events/page.tsx:38]: This is an example of how we would load completely different pages based on the locale.

src/components/app/localizedComponent.tsx
twistershark, line 1: This server component would help us change specific components on a shared page based on the user's locale.

An example of usage would be something like this:

```jsx
<LocalizedComponent  
  locale={locale}
  options={{
    'en-US': <div>Hello US</div>,
    'en-UK': <div>Hello UK</div>
  }} 
/>
```

src/intl/locales.ts
twistershark, line 1: Added UK as a new locale for the page example

Issue #1688, opened 2024-12-20
Fix initial load of the politicians page


lucasrmp-cb:
In the past the politicians page used to be rendered on the first page load, recently the table stopped rendering in the first load and it causes a giant layout shift. The objective here is to make sure the table loads on the first page render

PR #1702, opened 2024-12-24
Fix/politicians page initial render

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1688

fixes 

## What changed? Why?

- needed to add skeleton because the table now use dynamic API to have the filtering on the URL

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/dtsiClientPersonDataTable/dataTable.tsx
+++ b/src/components/app/dtsiClientPersonDataTable/dataTable.tsx
@@ -47,8 +47,6 @@ interface DataTableProps<TData extends Person = Person> extends Partial<TableOpt
   locale: SupportedLocale
 }
 
-type DataTableSkeletonProps<TData extends Person = Person> = Partial<TableOptions<TData>>
-
 export const SortableHeader = <TData extends Person = Person>({
   column,
   children,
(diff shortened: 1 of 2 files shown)


src/components/app/dtsiClientPersonDataTable/dataTable.tsx
```diff
@@ -221,3 +221,108 @@ export function DataTable<TData extends Person = Person>({
     </div>
   )
 }
+
+type DataTableSkeletonProps<TData extends Person = Person> = Partial<TableOptions<TData>>
+
+export function DataTableSkeleton<TData extends Person = Person>({
```
lucasrmp-cb, line 227: Great job on this, one question, is there a way to force a skeleton of the pagination buttons on the bottom? It's the only element that still causes layout shift on the page

lucasrmp-cb (changes requested):

src/components/app/dtsiClientPersonDataTable/dataTable.tsx
olavoparno: thanks man. oh, I think we can nice point

lucasrmp-cb (approved):

PR #1704, opened 2024-12-30
feat: builder.io

### LINKED ISSUE #1705: Builder.io initial setup
Relates to #1554

LucasdRossi:
closes #1705 

## What changed? Why?

- Basic logic for integration with Page Models
  - Logic to fetch the content
  - Component to render Builder.io blocks
- Webhook to revalidate page cache when a new version is published
- Changes to the CSP header to allow page previews in the Builder.io iframe only in the preview environment
- Logic to render the Navbar and Footer based on the page configuration in Builder.io

## Notes to reviewers

- There is only one dynamic route configured in Builder.io for testing, which can be accessed at <a href="https://swc-web-git-feat-builder-io-coinbase-vercel.vercel.app/content/test" target="_blank">/content/test</a>
- I am not certain about the file locations; I decided to place them inside `utils`. However, we could also create a `builder` folder at the root of the project, bringing together the two pieces of logic that are currently split between `/utils/web/builder` and `/utils/server/builder`

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/app/[locale]/layout.tsx
```diff
@@ -69,9 +70,13 @@ export default async function Layout({
         <TopLevelClientLogic locale={locale}>
           <FullHeight.Container>
             <NavBarGlobalBanner />
-            <Navbar locale={locale} />
+            <MaybeBuilderContent>
```
Comment (olavoparno) [inline src/app/[locale]/layout.tsx:73]: is this used to hide components if rendering a builder.io page?

src/components/ui/button/button.builder.tsx
```diff
@@ -0,0 +1,51 @@
+import { Builder, withChildren } from '@builder.io/react'
```
olavoparno, line 1: add "use client"

src/utils/server/builder/models/page/content.ts
```diff
@@ -0,0 +1,11 @@
+import { PageModel, PageModelNames } from '@/utils/server/builder/models/page/PageModel'
+
+const PAGE_PREFIX = '/content/'
+
+class ContentPageModel extends PageModel {
+  isBuilderPage(pathname: string): boolean {
+    return pathname.startsWith(this.routePrefix)
```
olavoparno, line 7: using `this` here is magic unless people really know the class it is extending

src/utils/server/builder/serverCMS.ts
```diff
@@ -2,10 +2,10 @@ import { builder } from '@builder.io/sdk'
 
 import { requiredOutsideLocalEnv } from '@/utils/shared/requiredEnv'
 
-const BUILDER_IO_PUBLIC_KEY = requiredOutsideLocalEnv(
-  process.env.BUILDER_IO_PUBLIC_KEY,
-  'BUILDER_IO_PUBLIC_KEY',
-  'all builder.io related',
+const NEXT_PUBLIC_BUILDER_IO_PUBLIC_KEY = requiredOutsideLocalEnv(
```
olavoparno, line 5: do we really need it to be publicly accessible? We should make this private

olavoparno, line 5: this could remain private, couldn't it? is there a need to expose it?

src/app/[locale]/topLevelClientLogic.tsx
```diff
@@ -1,5 +1,7 @@
 'use client'
 
+import '@/utils/web/builder/registerComponents'
```
Comment (travisbloom-cb) [inline src/app/[locale]/topLevelClientLogic.tsx:3]: we only want this imported on builder.io pages

src/components/app/builder.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type BuilderPageProps = ComponentProps<typeof BuilderComponent>
+
+maybeInitCMSClient()
+
+export function RenderBuilderContent(props: BuilderPageProps) {
+  const hasHydrated = useHasHydrated()
+  const isPreviewing = useIsPreviewing()
+
+  if (!hasHydrated) {
```
travisbloom-cb, line 18: Does this mean no builder.io pages will server side render? This seems like a pretty big issue/bug

src/components/app/builderPageLayout.tsx
```diff
@@ -0,0 +1,26 @@
...
+  pageModel: PageModel
+  children: React.ReactNode
+  locale: SupportedLocale
+  pathname: string
+}) {
+  const details = await pageModel.getPageMetadata(pathname)
+
+  return (
+    <>
+      {details.hasNavbar && <Navbar locale={locale} />}
```
travisbloom-cb, line 21: The better way of doing this is prob to have a Navbar and Footer builder component

src/components/ui/button/button.builder.tsx
```diff
@@ -0,0 +1,51 @@
...
+import dynamic from 'next/dynamic'
+
+import type { BuilderComponentBaseProps } from '@/utils/web/builder/types'
+
+import { buttonVariantsConfig } from '.'
+
+const Button = dynamic(() => import('@/components/ui/button').then(mod => mod.Button))
```
travisbloom-cb, line 8: Why are we making this a dynamic import?

src/utils/server/builder/models/data/events.ts
```diff
@@ -1,17 +1,58 @@
...
 
-import { builderIOClient } from '@/utils/server/builderIO/client'
+import { serverCMS } from '@/utils/server/builder/serverCMS'
 import { SWCEvents, zodEventSchemaValidation } from '@/utils/shared/getSWCEvents'
 import { getLogger } from '@/utils/shared/logger'
 import { NEXT_PUBLIC_ENVIRONMENT } from '@/utils/shared/sharedEnv'
 
+const logger = getLogger(`builderIOEvents`)
+
+export async function getEvent(eventSlug: string, state: string) {
```
travisbloom-cb, line 11: Why did we modify?

src/utils/server/builder/serverCMS.ts
LucasdRossi: Yes, we need to use it to initialize builder on the client [here](https://github.com/Stand-With-Crypto/swc-web/blob/d46412909af3bd17e01695d81bb2810caa5d46b2/src/utils/web/builder/clientCMS.ts), so it needs to be public

src/utils/server/builder/models/page/PageModel.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export interface PageMetadata {
+  title: string
+  description?: string
+  hasNavbar?: boolean
+  hasFooter?: boolean
+}
+
+export abstract class PageModel {
```
travisbloom-cb, line 17: We don't use class-based architecture anywhere else. Prob shouldnt start now

src/utils/server/builder/models/page/content.ts
```diff
@@ -0,0 +1,11 @@
+import { PageModel, PageModelNames } from '@/utils/server/builder/models/page/PageModel'
+
+const PAGE_PREFIX = '/content/'
+
+class ContentPageModel extends PageModel {
```
travisbloom-cb, line 5: see class comment

src/utils/server/builder/models/page/page.ts
```diff
@@ -0,0 +1,13 @@
+import { PageModel, PageModelNames } from '@/utils/server/builder/models/page/PageModel'
+
+class DefaultPageModel extends PageModel {
```
travisbloom-cb, line 3: see class comment

src/utils/server/builder/models/page/page.ts
```diff
@@ -0,0 +1,13 @@
...
+  public readonly pages = {
```
travisbloom-cb, line 4: This architecture is problematic. We don't want to have to update the codebase everytime someone needs to ship a new page. Can you explain why we need this?

src/utils/server/builder/models/page/utils/isBuilderPage.ts
```diff
@@ -0,0 +1,15 @@
+import { allPageModels } from '@/utils/server/builder/models/page'
+
+export function isBuilderPage(pathname: string | null) {
```
travisbloom-cb, line 3: see comment above. This architecture is problematic. We don't want to have to update the codebase everytime someone needs to ship a new page. Can you explain why we need this?

travisbloom-cb (changes requested):

src/app/[locale]/content/[[...page]]/page.tsx
```diff
@@ -0,0 +1,53 @@
+import { Metadata } from 'next'
+
+import { RenderBuilderContent } from '@/components/app/builder'
+import { BuilderPageLayout } from '@/components/app/builderPageLayout'
+import { PageProps } from '@/types'
+import { contentPageModel } from '@/utils/server/builder/models/page'
+import { serverCMS } from '@/utils/server/builder/serverCMS'
```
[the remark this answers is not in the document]
Comment (LucasdRossi) [reply src/app/[locale]/content/[[...page]]/page.tsx:7]: I agree. Do you think we should keep it inside the `utils` folder or have a `builderio` folder directly inside `src`?

src/utils/server/builder/models/data/events.ts
LucasdRossi: I didn't modify the function, just moved it to a new location and grouped all of the events data model fetching logic to the same file

src/components/ui/button/button.builder.tsx
LucasdRossi: You're right, we should only use dynamic imports for non-critical components. I'll remove the dynamic import

src/app/[locale]/content/[[...page]]/page.tsx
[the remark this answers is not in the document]
Comment (travisbloom-cb) [reply src/app/[locale]/content/[[...page]]/page.tsx:7]: i think the folder structure is fine, was mainly commenting on the variable name

src/components/ui/button/button.builder.tsx
travisbloom-cb: I dont think we'd want any dynamic imports. If there are non critical components needed for a one-off builder page, wed prob just add the page directly to the codebase and import the registration

src/utils/server/builder/models/page/utils/getPageContent.ts
```diff
@@ -0,0 +1,17 @@
+import { builderSDKClient } from '@/utils/server/builder/builderSDKClient'
+import { PageModelIdentifiers } from '@/utils/server/builder/models/page/constants'
+
+export function getPageContent(pageModelName: PageModelIdentifiers, pathname: string) {
```
olavoparno, line 4: should we turn this function into async?

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return generateMetadataDetails({
+    title: metadata.title,
+    description: metadata.description,
+  })
+}
+
+export async function generateStaticParams() {
+  const paths = await builderSDKClient
+    .getAll(PAGE_MODEL, { options: { noTargeting: true } })
```
Comment (travisbloom-cb) [inline src/app/(builder)/[locale]/content/[[...page]]/page.tsx:48]: This will need to be changed as soon as we have builder.io pages that arent prefixed with /content/. Can't we just filter for builder.io pages that start with /content/?

src/app/(builder)/[locale]/layout.tsx
```diff
@@ -0,0 +1,82 @@
+import { SpeedInsights } from '@vercel/speed-insights/next'
```
Comment (travisbloom-cb) [inline src/app/(builder)/[locale]/layout.tsx:1]: smart usage of next.js folder apis 👍

src/components/app/builder.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function RenderBuilderContent(props: BuilderPageProps) {
+  const isPreviewing = useIsPreviewing()
+
+  if (props.content || isPreviewing) {
+    return <BuilderComponent {...props} />
+  }
+
+  if (props.type === 'section') {
```
travisbloom-cb, line 18: is this still needed?

src/components/app/builder.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return null
+  }
+
+  return notFound()
```
travisbloom-cb, line 22: this should prob log to sentry?

src/utils/server/builder/models/data/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum DataModelIdentifiers {
```
travisbloom-cb, line 1: `BuilderDataModelIdentifiers`

olavoparno (dismissed):

src/utils/server/builder/models/page/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum PageModelIdentifiers {
```
travisbloom-cb, line 1: `BuilderPageModelIdentifiers`

src/utils/server/builder/models/page/utils/getPageContent.ts
```diff
@@ -0,0 +1,17 @@
...
+  return builderSDKClient
+    .get(pageModelName, {
+      userAttributes: {
+        urlPath: pathname,
+      },
+      // Set prerender to false to return JSON instead of HTML
+      prerender: false,
+      // Set cachebust to true to get the latest content.
+      // This should only be used to generate static pages.
+      cachebust: true,
```
travisbloom-cb, line 14: when would this be false?

src/utils/server/builder/models/page/utils/getPageDetails.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function getPageDetails(
+  pageModelName: PageModelIdentifiers,
+  pathname: string,
+): Promise<PageMetadata> {
+  const content = await getPageContent(pageModelName, pathname)
+
+  return {
+    title: content?.data?.title,
+    description: content?.data?.description,
+    hasFooter: content?.data?.hasFooter ?? true,
```
travisbloom-cb, line 20: can we set this to default within builder.io UI? Confusing it could be unchecked in UI but true here

travisbloom-cb (approved):

src/components/app/builder.tsx
lucasrmp-cb: If this uses the default not found page it already handles logging and mixpanel events

src/utils/server/builder/models/page/constants.ts
twistershark: +1

src/utils/server/builder/models/data/constants.ts
twistershark: +1

src/utils/server/builder/models/page/utils/getPageDetails.ts
```diff
@@ -0,0 +1,23 @@
...
+    hasNavbar: content?.data?.hasNavbar ?? true,
```
lucasrmp-cb, line 21: Same here

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
Comment (twistershark) [inline src/app/(builder)/[locale]/content/[[...page]]/page.tsx:1]: What is your opinion on having the builder io group folder inside our current [locale] folder instead of creating another folder at the root level?

it would be something like 

[locale]
  (builder)
    content
    ...

Comment (twistershark) [reply src/app/(builder)/[locale]/content/[[...page]]/page.tsx:1]: Also, I'm not sure if we want to keep the "/content" segment on Builder.io pages, but I think we could move the [[...page]] folder to the same level as our other pages. NextJS will prioritize loading content from non-dynamic folders, and if none is found, it will fallback to dynamic folders (routes) if they exist. 

Since you are checking if the page exists in Builder.io, we would still show 404 pages when it doesn’t exist.

src/utils/server/builder/models/page/utils/getPageDetails.ts
LucasdRossi: Yes, the default value is `true` in Builder.io, this `?? true` will only be applied if `content?.data?.hasFooter` is `undefined`

image

If the checkbox is unchecked, this value will be false

src/utils/server/builder/models/page/utils/getPageContent.ts
LucasdRossi: I don't think we should, because we don't need to await for anything to return the content

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
Comment (LucasdRossi) [reply src/app/(builder)/[locale]/content/[[...page]]/page.tsx:48]: This is only getting pages within the `content` page model, so it's already fetching pages that start with `/content/` because `PAGE_MODEL = content`.

When we add pages to the default page model they won't be fetched here, because they belong to another page model

src/utils/server/builder/models/page/utils/getPageContent.ts
LucasdRossi: This is what the [official doc](https://www.builder.io/c/docs/content-api#code-cachebust-code) says
image

src/components/app/builder.tsx
LucasdRossi: It's using the default not found page, but do you guys think we should log a different type of error?

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
```diff
@@ -0,0 +1,58 @@
...
+    .then(res => res.map(({ data }) => data?.url))
+
+  return paths.map((path: string) => {
+    return {
+      params: {
+        page: path.replace(PAGE_PREFIX, '').split('/'),
```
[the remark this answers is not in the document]
Comment (LucasdRossi) [reply src/app/(builder)/[locale]/content/[[...page]]/page.tsx:54]: The Builder.io response already includes the page prefix, but Next.js only requires the slug (without the prefix) to correctly generate static pages. That's why the path is adjusted here.

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
Comment (LucasdRossi) [reply src/app/(builder)/[locale]/content/[[...page]]/page.tsx:1]: Thank you for the great feedback, Paulo!

**Regarding the route grouping:**
We grouped the pages this way because Builder.io pages need to control whether the Navbar and Footer are displayed based on their configuration. To achieve this, we require the page’s pathname, which is not accessible in the Layout. Hence, we introduced the BuilderPageLayout component as a top-level component for Builder.io pages. This component registers and renders the Navbar and Footer conditionally, avoiding duplication. The Builder.io folder layout does not include fixed Navbar and Footer to preserve the existing layout structure without modifying all pages to add separate top-level components for these elements.

**Regarding removing the /content prefix:**
This is a valid approach, but it limits control over dynamic entries in the application. Without the prefix, managing content in Builder.io becomes more challenging, as we’d have to rely on a single, general Page Model. By keeping defined dynamic points like /blog, /content, or /press, each with its own Page Model in Builder.io, we improve maintainability and allow for specific configurations for each type of page.

Let me know your thoughts!

olavoparno (dismissed):

src/app/(builder)/[locale]/content/[[...page]]/page.tsx
Comment (twistershark) [reply src/app/(builder)/[locale]/content/[[...page]]/page.tsx:1]: Got it! Great work!

twistershark (dismissed):

lucasrmp-cb (dismissed):

lucasrmp-cb (approved):

Issue #1697, opened 2024-12-21
Follow up after Next 15


olavoparno:
upgrade the rest of the dependencies after Next 15 upgrade

olavoparno:
we're waiting a resolution on the [cypress failing on main issue ](https://github.com/Stand-With-Crypto/swc-web/issues/1715) because both of them involves lots of changes in package-lock/package.json. The idea is to avoid conflicts.

PR #1716, opened 2025-01-06
Chore/upgrade next 15 follow up

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1697

fixes 

## What changed? Why?

Upgraded libs so we get rid of overrides. We still have to override libs for react-simple-maps due to lack of support on their side.

Also, due to upgrading react-email, we had API changes as well.

## How has it been tested?

- [X] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


package.json
```diff
@@ -197,9 +197,9 @@
...
   },
   "overrides": {
-    "@types/react": "^19.0.1",
-    "@types/react-dom": "^19.0.2",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0"
+    "react-simple-maps": {
+      "react": "^19.0.0",
+      "react-dom": "^19.0.0"
+    }
```
lucasrmp-cb, line 203: Can we add context on why?

src/utils/server/email/templates/reactivationReminder.tsx
```diff
@@ -44,15 +44,14 @@ export default function ReactivationReminder({
...
-          </ul>
         </Text>
-        <br />
+        <ul className="pl-4 text-[15px] text-[#5B616E]">
+          <li>Job creation</li>
+          <li>Technology and innovation</li>
+          <li>Modernizing the financial system</li>
+          <li>Protecting and empowering consumers</li>
+          <li>Maintaining our edge over other nations</li>
+        </ul>
```
lucasrmp-cb, line 54: Did we test that this doesn't change UI? We're removing all styles of `Text` from this section

src/utils/server/email/templates/becomeMemberReminder.tsx
```diff
@@ -48,25 +48,26 @@ export default function BecomeMemberReminderEmail({
...
+          <li>
+            Election analysis from top crypto political strategists who are making big moves leading
+            into November
+          </li>
+          <li>More chances to connect with fellow SWC members - particularly other full members</li>
+          <li>
+            Briefings from major crypto executives on how they see the political landscape and the
+            future of crypto in America
+          </li>
+        </ul>
```
lucasrmp-cb, line 68: Did we test that this doesn't change UI? We're removing all styles of Text from this section

lucasrmp-cb (changes requested):

package.json
olavoparno: simply because react-simple-maps does not have support yet :(

src/utils/server/email/templates/reactivationReminder.tsx
olavoparno: for some reason it wouldn't render and would break if left the way it was, so like this we didn't have any major UI change

src/utils/server/email/templates/becomeMemberReminder.tsx
olavoparno: Had to be done due to the upgrade. No major changes in the UI though

image

lucasrmp-cb (approved):

PR #1728, opened 2025-01-08
feat: restore popover behavior for the address field

### LINKED ISSUE #1727: Restore the combobox functionality for the address field
Because of the new congress email campaign, we had to temporary change the address field behavior on desktop to be a dialog instead of a popover.

We need to find the issue and restore the default behavior for a better UX

twistershark:
closes #1727 
## What changed? Why?

This PR restores the combobox behavior for the address field and removes the temporary dialog behavior.

Based on this issue: https://github.com/radix-ui/primitives/issues/2121, the bug was caused by an outdated `@radix-ui/react-dismissable-layer` version that was used on the dialog and popover components. By upgrading them to the latest `v1.1.4` this peer dependency is also upgraded to the latest version, fixing the non clickable bug.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


package.json
```diff
@@ -48,11 +48,11 @@
     "@radix-ui/react-accordion": "^1.1.2",
     "@radix-ui/react-checkbox": "^1.0.4",
     "@radix-ui/react-collapsible": "^1.0.3",
-    "@radix-ui/react-dialog": "^1.1.2",
+    "@radix-ui/react-dialog": "1.1.4",
```
olavoparno, line 51: should it be fixed on this version or should we keep the ˆ marker?

twistershark: Thought about making it fixed to prevent unexpected bugs in the future

LucasdRossi (approved):

Issue #1740, opened 2025-01-10
Wrong field targeting on browser autofill


lucasrmp-cb:
User Report: I was going through the "email your congress member" flow and my browser's auto-fill form accidentally overrode the prefilled email body with my address instead. I didn't even realize it until I finished entering my info to find the body was overridden. Seems like there's some incorrect form targeting going on here.

![Image](https://github.com/user-attachments/assets/9546bebd-6102-40dc-b5d2-054e89f62085)
![Image](https://github.com/user-attachments/assets/4c500122-566e-47ec-8821-235074688642)
![Image](https://github.com/user-attachments/assets/417076b0-f9fd-4c2c-9c0a-a0df825dc0ac)

lucasrmp-cb:
This happened in Brave, but most likely will happen in other chromium based browsers as well

PR #1743, opened 2025-01-13
feat: add autoComplete off attribute to message textarea component

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1740

fixes 

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/userActionFormEmailCongressperson/index.tsx
+++ b/src/components/app/userActionFormEmailCongressperson/index.tsx
@@ -329,6 +329,7 @@ export function UserActionFormEmailCongressperson({
                       )}
                       <FormControl>
                         <Textarea
+                          autoComplete="off"
                           placeholder=""
                           rows={16}
                           {...field}
--- a/src/components/app/userActionFormEmailCongressperson/index.tsx
+++ b/src/components/app/userActionFormEmailCongressperson/index.tsx
@@ -330,8 +330,10 @@ export function UserActionFormEmailCongressperson({
                       <FormControl>
                         <Textarea
                           autoComplete="off"
+                          autoCorrect="off"
                           placeholder=""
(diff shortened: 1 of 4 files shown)


src/components/app/userActionFormEmailCongressperson/index.tsx
```diff
@@ -316,7 +316,7 @@ export function UserActionFormEmailCongressperson({
               />
               <FormField
                 control={form.control}
-                name="message"
+                name="contactMessage"
```
twistershark, line 319: the field having the name "message" was a problem? Wasn't expecting it, but makes a lot of sense.

twistershark (approved):

src/components/app/userActionFormEmailCongressperson/index.tsx
olavoparno: this is to trick browsers to not follow their cached autofill, hopefully

src/components/app/userActionFormEmailCongressperson/index.tsx
```diff
@@ -329,8 +329,11 @@ export function UserActionFormEmailCongressperson({
                       )}
                       <FormControl>
                         <Textarea
+                          autoComplete="off"
+                          autoCorrect="off"
                           placeholder=""
                           rows={16}
+                          spellCheck={false}
```
lucasrmp-cb, line 336: Just doing this didn't stop the behavior? Is the name change necessary?

olavoparno: what do you think about it @lucasrmp-cb?

Issue #1748, opened 2025-01-15
Flag specific field validation errors as systematic and log on Sentry


lucasrmp-cb:
We had a bug recently where a field validation for the `dtsiSlug` field was failing on the email action, this was caused by a problem with manually inputed politicians on DTSI, since we treat all field validation errors as user input failures, this was just logged as an info and ignored.

The problem is that this isn't a user inputted value so if that validation fails we should treat is as a system failure and log it appropriately. The idea here is to create an api where we tag non-user-inputted fields and throw the error as a system failure if that validation fails

lucasrmp-cb:
This thread has some ideas we might wanna leverage
https://github.com/colinhacks/zod/issues/1718

PR #1750, opened 2025-01-15
feat: enhance validation error handling with metadata in user action …

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1748

fixes 

## What changed? Why?

- Added Refactored Zod validation to utilize `withSafeParseWithMetadata` and `withEnhancedDescription` for improved error metadata extraction.
- Updated `_actionCreateUserActionEmailCongressperson` to include error metadata in the response when validation fails.
- Modified `triggerServerActionForForm` to handle and log error metadata with Sentry for improved error tracking.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/validation/forms/zodUserActionFormEmailCongressperson.ts
```diff
@@ -13,7 +13,14 @@ const base = object({
...
   subject: string().trim(),
-  dtsiSlugs: array(zodDTSISlug).min(1),
+  dtsiSlugs: array(zodDTSISlug)
+    .min(1)
+    .describe(
+      JSON.stringify({
+        triggerException: true,
+        message: 'Something went wrong with the DTSI Slug. Please check against database.',
+      }),
+    ),
```
lucasrmp-cb, line 23: I would abstract this describe logic into a wrapper, would prefer an api that looks something like this
```suggestion
  dtsiSlugs: withMetadata(
      array(zodDTSISlug).min(1),
      {
        triggerException: true,
        message: 'Something went wrong with the DTSI Slug. Please check against database.',
      },
    ),
```

lucasrmp-cb: Idea being that we'd have another function called `parseMetadata` or `getSchemaMetadata` that would work as a parse and that way all this description and `JSON.stringify` logic would be encapsulated

src/validation/utils.ts
```diff
@@ -1,4 +1,52 @@
...
+        const metadataDescription =
+          zodSchema instanceof z.ZodObject
+            ? zodSchema.shape[field as keyof typeof zodSchema.shape].description
+            : undefined
+        const parsedMetadataDescription =
+          typeof metadataDescription === 'string' &&
+          metadataDescription.startsWith('{') &&
+          metadataDescription.endsWith('}')
+            ? JSON.parse(metadataDescription)
+            : undefined
```
lucasrmp-cb, line 37: And that `parseMetadata` util function would have this logic

src/utils/web/formUtils.ts
```diff
@@ -21,7 +21,21 @@ export type GenericErrorFormValues = {
...
+  Fn extends (args: P) => Promise<
+    | {
+        errors: Record<string, string[]>
+        errorsMetadata?: {
+          field: string
+          data: {
+            triggerException: boolean
+            severityLevel?: SeverityLevel
+            message?: string
+          }
```
lucasrmp-cb, line 33: My understand is that if `triggerException` is `false`, we won't have `severityLevel` and `message`, if that's the case we could change this type to reflect that better
```suggestion
          data: {
            triggerException: false
            severityLevel?: never
            message?: never
          } | {
            triggerException: true
            severityLevel: SeverityLevel
            message: string
          }
```

Doing it like this would make both fields required when you set `triggerException` as true

src/utils/web/formUtils.ts
```diff
@@ -96,12 +110,33 @@ export async function triggerServerActionForForm<
...
+
+      if ('errorsMetadata' in response) {
+        response.errorsMetadata?.forEach(errorMetadata => {
+          const { field, data } = errorMetadata
+          const { triggerException, severityLevel, message } = data
+
+          if (triggerException) {
+            Sentry.captureException(message ?? response.errors[field]?.join('. ') ?? '', {
+              level: severityLevel ?? 'error',
+              extra: {
```
lucasrmp-cb, line 122: Let's include the field name here

lucasrmp-cb (changes requested):

src/validation/forms/zodUserActionFormEmailCongressperson.ts
olavoparno: I tried doing something like that but since this is all chainable from zod's, I don't think we can do much here you know? any thoughts?

src/utils/web/formUtils.ts
olavoparno: thanks man, really appreciate it

src/utils/shared/zod/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+function safeParseDescription({
+  description,
+  additionalMetadata,
+}: SafeParseDescriptionWithMetadata): string {
+  try {
+    return JSON.stringify(Object.assign({}, description, additionalMetadata))
+  } catch (error) {
+    return 'FAILED_TO_PARSE_DESCRIPTION'
+  }
```
lucasrmp-cb, line 36: I'm aligned with a graceful degradation in case this fails, but let's add a Sentry log to keep track, this would only happen on an eng mistake if I understood correctly

src/validation/forms/zodUserActionFormEmailCongressperson.ts
```diff
@@ -13,7 +14,10 @@ const base = object({
     .min(1, 'Please enter a message')
     .max(2000, 'Your message should not exceed 2000 characters'),
   subject: string().trim(),
-  dtsiSlugs: array(zodDTSISlug).min(1),
+  dtsiSlugs: withEnhancedDescription(array(zodDTSISlug).min(1), {
+    triggerException: true,
+    message: 'Something went wrong with the DTSI Slug. Please check against database.',
```
lucasrmp-cb, line 19: This is a user facing error, let's use a [generic error message](https://github.com/Stand-With-Crypto/swc-web/blob/b1fed71c2624e397ce091dad2df5916fdedd95bd/src/utils/web/errorUtils.ts#L1)

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

PR #1755, opened 2025-01-16
Fix/logout impossible state

### LINKED ISSUE #1753: Create refresh token logic to thirdweb auth
We identified a case where if the thirdweb auth token is expired we get into a state where the login button is stuck and doesn't work because the `/identified-user/full-profile-info` route would return null even though thirdweb was connected, so thirdweb would think the user is logged in while we would think the user is not.

To solve this issue we should add a refresh token logic to the app, where we use the `refreshJWT` function from `thirdweb/utils` to generate another jwt everytime the client requests the user information with an expired jwt

lucasrmp-cb:
closes #1753 

## What changed? Why?

This adds a token refresh logic to the `appRouterGetThirdwebAuthUser` util. Besides that it adds some debugging tools to the internal user settings pages.

The issue here is that expired jwt tokens made so the user got stuck by thirdweb thinking it was authenticated while we thought the user was

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/thirdweb/onLogout.ts
```diff
@@ -12,15 +13,23 @@ export async function onLogout() {
   const currentCookies = await cookies()
   const localUser = await parseLocalUserFromCookies()
 
-  const token = currentCookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)
-  const decodedToken = token?.value ? jwtDecode<{ ctx?: { userId?: string } }>(token.value) : null
-  const { userId } = decodedToken?.ctx ?? {}
+  try {
+    const token = currentCookies.get(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX)
+    const decodedToken = token?.value ? jwtDecode<{ ctx?: { userId?: string } }>(token.value) : null
```
lucasrmp-cb, line 18: `jwtDecode` might throw on invalid token, just doing this to make sure we delete the cookie either way

olavoparno (approved):

PR #1759, opened 2025-01-20
feat: migrate Privacy Policy page to Builder.io

### LINKED ISSUE #1707: Move /privacy page to Builder.io
relates to #1554

LucasdRossi:
closes #1707 

## What changed? Why?

Moved the /privacy page to Builder.io. You can preview the page [here](https://swc-web-git-feat-builder-privacy-page-coinbase-vercel.vercel.app/privacy)

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


olavoparno (dismissed):

src/components/app/builder/builderComponent.tsx
```diff
@@ -4,13 +4,22 @@ import type { ComponentProps } from 'react'
 import { Builder, BuilderComponent, useIsPreviewing } from '@builder.io/react'
 import { notFound } from 'next/navigation'
 
+import { useSession } from '@/hooks/useSession'
+
 type BuilderPageProps = ComponentProps<typeof BuilderComponent>
 
 export function RenderBuilderContent(props: BuilderPageProps) {
+  const session = useSession()
```
travisbloom-cb, line 12: Is there a way to make these components react server components? Or is that not supported by builder?

src/components/ui/button/button.builder.tsx
```diff
@@ -1,22 +1,69 @@
...
+      } else if (type === 'internal') {
+        Comp = (
+          <Button {...buttonProps} asChild>
+            <InternalLink href={href}>{props.children}</InternalLink>
+          </Button>
+        )
+      }
+    }
+
+    if (props.requireAuthentication) {
```
travisbloom-cb, line 47: we might want a seperate component for this? Im fine either way but this seems like functionality thats beyond a base component like button

src/utils/web/builder/components/richText.builder.tsx
```diff
@@ -0,0 +1,61 @@
...
+Builder.registerComponent(
+  ({ text, attributes }: BuilderTextProps) => (
+    // Replace text-indent with margin-left because Builder.io applies text-indent to nested lists and
+    // RichText uses tailwind typography prose class which doesn't apply text-ident to the ::marker pseudo-element
+    // So we need to use margin-left instead
+    <RichText content={text} {...sanitizeBuilderAttributes(attributes)} />
+  ),
+  {
+    name: 'Rich Text',
+    noWrap: true, // Disables the default "Link URL" field
```
travisbloom-cb, line 51: doesnt this component also override a default?

src/utils/web/builder/sanitizeBuilderAttributes.ts
```diff
@@ -0,0 +1,7 @@
+import { BuilderComponentAttributes } from '@/utils/web/builder/types'
+
+export function sanitizeBuilderAttributes(attributes?: BuilderComponentAttributes) {
```
travisbloom-cb, line 3: can we get some commentary about why?

travisbloom-cb:
image

src/utils/web/builder/components/richText.builder.tsx
LucasdRossi: No, there's not a default component called `Rich Text`, but we're overriding the default componente `Text` and only allowing it to use plain text, this way we can use it inside buttons and other components.

src/components/app/builder/builderComponent.tsx
LucasdRossi: Unfortunately it's not supported yet. I found a [thread](https://forum.builder.io/t/ssr-and-ssg-on-next-js-with-approuter/4993) on their forum talking about this and they said that it's only supported in their [Builder NextJS SDK](https://www.npmjs.com/package/@builder.io/sdk-react-nextjs), which is in beta

olavoparno (dismissed):

src/utils/web/builder/sanitizeBuilderAttributes.ts
LucasdRossi: I was using this to remove the `key` from the `attributes` prop because there was an error saying that this prop shouldn't be spreaded, but explicitly passing the key resolved the issue so I removed this function

```jsx
// before
<Text text={text} {...attributes} />


// after
<Text text={text} {...attributes} key={attributes?.key} />
```

src/components/ui/button/button.builder.tsx
LucasdRossi: Created a component called `RequireAuthentication` and moved this logic to Builder.io

olavoparno (dismissed):

src/components/app/builder/builderComponent.tsx
LucasdRossi: Builder content is still being rendered in the server

image

olavoparno (approved):

src/utils/web/builder/components/richText.builder.tsx
travisbloom-cb: Wait, so builder.io has a text component that already allows rich text, and we're overriding it to not allow rich text, and creating a seperate rich text component?

Why don't we just override text directly? Why do we need two components?

src/utils/web/builder/editorSettings.ts
```diff
@@ -0,0 +1,12 @@
+import { Builder } from '@builder.io/react'
+
+import { openSansFont, satoshiFont } from '@/utils/web/fonts'
+
+Builder.register('editor.settings', {
+  designTokens: {
```
travisbloom-cb, line 6: 👀

src/utils/web/builder/components/richText.builder.tsx
LucasdRossi: image

This is the main reason why I've decided to go with two separate components: The Rich Text component has the `prose` class which applies some tailwind typography styling to the text as you can see in the left button, but the text component has no default styling so it can inherit it's parent styling, so if we change the variant of the button in the right, the color of the text would also change accordingly to the variant, when changing the variant in the left button wouldn't have the same result.

lucasrmp-cb (approved):

olavoparno (approved):

PR #1772, opened 2025-01-23
fix: remove app dependency from mocks

### LINKED ISSUE #1715: Cypress failiing on main
https://github.com/Stand-With-Crypto/swc-web/actions/runs/12633266917/job/35198537059

lucasrmp-cb:
closes #1715
opens #1773 

## What changed? Why?

We started experiencing cypress webpack build errors where it complained about a `next/router` not being explicit (`next/router.js`) while compiling. After considering and trying a few solutions changing the build system setup for cypress, sentry and nextjs, we came to the realization that the error wasn't on the app, was on the cypress task runner and that meant that `sentry/nextjs` and therefore `next` was being compiled as a dependency of cypress, that should never happen... Cypress should just run test scripts against an app build and be completely ignorant of internal libraries that app uses internally (like `sentry/nextjs` and `next`).

After that realization I started looking for import statements done inside of `.cy.ts` files that would be included in a compilation step as a dependency, and in one test file I found the import of a phone number validation function that in it's definition file imported sentry, that was the issue.

So the technical problem was that we were importing `@sentry/nextjs` in a file that was a compilation dependency for cypress and (now I'm guessing) in some unidentified library (or codependency) update the webpack configuration for cypress stopped being able to compile sentry.

Also I see an architectural issue here where our e2e tests had (and still have) dependencies from the domain of the application (`src/` code is being imported from `cypress/`), as said previously this shouldn't be done since e2e tests should be treated as a stand alone project (even when sharing a repo) so the test is actually consistent. For example the fact we were using the app's phone number validation code to generate a phone number means we could never test an invalid phone number with the mocking interface we created and that's not clear for engineers. Also that meant we, in some instances, were registering Sentry logs for invalid phone numbers being generated (even though it was in preview it didn't make sense to log).

This PR is a solution for the critical issue, removing the dependency that added Sentry to the compilation and also is fixing all tests we've broken while cypress wasn't working. As a follow up to this we're gonna do a rework to the cypress setup where we will block any imports from `src/` to be done from `cypress/` and this will prevent a dependency to be created and ensure the independency of the app and e2e projects.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/mocks/mockValidPhoneNumber.ts
```diff
@@ -1,6 +1,6 @@
 import { faker } from '@faker-js/faker'
-
-import { validatePhoneNumber } from '@/utils/shared/phoneNumber'
```
lucasrmp-cb, line 3: Removing this import is the real fix, the rest is fixing tests we've broken while cypress was erroring

twistershark (approved):

Issue #1712, opened 2025-01-03
Move /partners page to Builder.io


LucasdRossi:

LucasdRossi:
[/partners](https://www.standwithcrypto.org/partners)

PR #1800, opened 2025-01-30
feat: add builder.io partners page

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1712

fixes 

## What changed? Why?

- added title, subtitle and partners page
- added partners structured data model on builder.io 

quick note: /partner page is some kind of hybrid implementation which renders content from builder.io page and next.js unregistered components with fetch data from builder.io structured data model.

## How has it been tested?

- [X] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/builder/builderComponent.tsx
```diff
@@ -4,12 +4,14 @@ import type { ComponentProps } from 'react'
...
 import { notFound } from 'next/navigation'
 
+import { useHasHydrated } from '@/hooks/useHasHydrated'
 import { useSession } from '@/hooks/useSession'
 import { BuilderState } from '@/utils/web/builder/types'
 
 type BuilderPageProps = ComponentProps<typeof BuilderComponent>
 
 export function RenderBuilderContent(props: BuilderPageProps) {
+  const hasHydrated = useHasHydrated()
```
olavoparno, line 14: We should discuss about this again @LucasdRossi @lucasrmp-cb since we're already at a client component, let's bring up why we should/should not use `useHasHydrated` here.

lucasrmp-cb: You should never use a hydration check to render layout that is part of the initial page load UI. This always causes layout shift and it's actually a non-solution for the hydration warning

lucasrmp-cb: With this change you made so this is rendered to the client, and it should never be the case. And that's not counting that this completely disables SEO for the content coming from builder io
image

lucasrmp-cb (changes requested):
Gonna block merge pending the client side rendered problem

src/components/app/builder/builderComponent.tsx
lucasrmp-cb: And to address the "since we're already at a client component", Next still do SSR on client components to html, that's why the initial load has the html pre rendered
https://github.com/reactwg/server-components/discussions/4

olavoparno: agreed, just shouting out for ideas and thoughts because we're facing this issue now

next.config.ts
```diff
@@ -231,6 +231,7 @@ const nextConfig: NextConfig = {
         hostname: 'cdn.builder.io',
       },
     ],
+    dangerouslyAllowSVG: true,
```
LucasdRossi, line 234: 👀

src/components/ui/pageSubTitle/pageSubTitle.builder.tsx
```diff
@@ -0,0 +1,55 @@
...
+        type: 'string',
+        required: true,
+        defaultValue: 'Enter some text...',
+      },
+      {
+        name: 'as',
+        type: 'enum',
+        defaultValue: 'h2',
+        enum: AsVariantsConfig,
+      },
```
LucasdRossi, line 46: I don't think we need this input, because non-technical people don't know about HTML heading elements

src/components/ui/pageTitleText/pageTitleText.builder.tsx
```diff
@@ -0,0 +1,50 @@
...
+        type: 'string',
+        required: true,
+        defaultValue: 'Enter some text...',
+      },
+      {
+        name: 'as',
+        type: 'enum',
+        defaultValue: 'h1',
+        enum: AsVariantsConfig,
+      },
```
LucasdRossi, line 41: I don't think we need this input, because non-technical people don't know about HTML heading elements

src/utils/server/builder/models/data/partners.ts
```diff
@@ -0,0 +1,62 @@
...
+      .filter(Boolean) as SWCPartners
+
+    if (filteredIncompletePartners.length === 0) return null
+
+    return filteredIncompletePartners
+  } catch (e) {
+    logger.error('error getting all partners:' + e)
+    Sentry.captureException(e, {
+      level: 'error',
+      tags: { domain: 'getPartners' },
```
LucasdRossi, line 58: Let's use the same domain as other builder.io errors for better filtering: `builder.io`

src/components/app/pagePartners/partnerGrid.tsx
```diff
@@ -1,22 +1,50 @@
...
+      imageLink: {
+        Default: partner.url,
+      },
+      imageAlt: partner.title,
+    }
+  })
+}
+
+export function PartnerGrid({ partners }: { partners: SWCPartners | null }) {
+  const currentPartners = mapPartnerWithFallback(partners || INDUSTRY_PARTNERS)
```
LucasdRossi, line 29: I don't think we should keep this `INDUSTRY_PARTNERS` in the project, because then we need to update two places if the list changes

src/components/app/builder/builderComponent.tsx
olavoparno: this isnt the case anymore, now items are being fetch differently so no hydration warnings

next.config.ts
olavoparno: removed, thanks for spotting

src/components/app/pagePartners/partnerGrid.tsx
olavoparno: this is because `INDUSTRY_PARTNERS` is used at home too, so if we removed this we would have to fetch at home too

src/utils/server/builder/models/data/partners.ts
olavoparno: we had this in events, but I agree, let me change everywhere we can

src/components/ui/pageTitleText/pageTitleText.builder.tsx
olavoparno: as we discussed, I've added it under advanced props. thanks man

src/app/api/public/builder/partners/data/route.ts
```diff
@@ -0,0 +1,45 @@
...
+      },
+      tags: {
+        domain: 'builder.io',
+      },
+    })
+
+    return new NextResponse('Unauthorized', {
+      status: 401,
+    })
+  }
```
lucasrmp-cb, line 35: Let's create an authentication middlware (hof would work) to handle authentication for builder io

src/components/app/pagePartners/partnerGrid.tsx
lucasrmp-cb: I agree with Lucas B here, we should fetch partners on the home page, keeping this here takes away from the "not having eng work to change" mentality

lucasrmp-cb (changes requested):

src/components/app/pagePartners/partnerGrid.tsx
olavoparno: we should add things to the home page with care because of its high access rate but ok, I'm adding it

src/components/app/pagePartners/partnerGrid.tsx
```diff
@@ -1,22 +1,25 @@
...
+export function PartnerGrid({ partners }: { partners: SWCPartners | null }) {
   return (
     <div className="flex flex-row flex-wrap items-center justify-center gap-3 sm:gap-8">
-      {INDUSTRY_PARTNERS.map(({ title, url, imageSrc }) => {
+      {partners?.map(currentParter => {
+        const { name, image, imageLink, imageAlt } = currentParter.data
+
         return (
-          <ExternalLink href={url} key={title}>
+          <ExternalLink href={imageLink.Default} key={name}>
```
LucasdRossi, line 12: I saw that this field is using the Localize feature from Builder.io, but this is not currently supported by our plan, so we shouldn't use it

src/components/ui/pageTitleText/pageTitleText.builder.tsx
```diff
@@ -0,0 +1,51 @@
...
+      ...props.attributes,
+      as: props.as,
+      size: props.size,
+    }
+    return (
+      <PageTitle {...pageTitleProps} key={props.attributes?.key} size={props.size}>
+        {props.title}
+      </PageTitle>
+    )
+  },
```
LucasdRossi, line 24: NP: you're passing `size` two times to the component, either you should use the `pageTitleProps` or pass each prop to the component like the `key` and `size` and spread the `attributes` prop.

Personally I prefer the second approach as you're only instantiating the component one time

src/components/ui/pageSubTitle/pageSubTitle.builder.tsx
```diff
@@ -0,0 +1,56 @@
...
+      as: props.as,
+      size: props.size,
+    }
+
+    return (
+      <PageSubTitle {...pageSubTitleProps} key={props.attributes?.key} size={props.size}>
+        {props.title}
+      </PageSubTitle>
+    )
+  },
```
LucasdRossi, line 29: Same NP as the pageTitleText component

src/components/app/pagePartners/partnerGrid.tsx
```diff
@@ -1,22 +1,25 @@
...
 import { ExternalLink } from '@/components/ui/link'
-import { INDUSTRY_PARTNERS } from '@/utils/shared/industryPartners'
+import { SWCPartners } from '@/utils/shared/getSWCPartners'
 
-export function PartnerGrid() {
+export function PartnerGrid({ partners }: { partners: SWCPartners | null }) {
   return (
     <div className="flex flex-row flex-wrap items-center justify-center gap-3 sm:gap-8">
-      {INDUSTRY_PARTNERS.map(({ title, url, imageSrc }) => {
+      {partners?.map(currentParter => {
```
LucasdRossi, line 8: There is a typo here, it should be `currentPartner`

src/components/app/pagePartners/partnerGrid.tsx
olavoparno: thanks man, i've removed

src/app/api/public/builder/events/page/route.ts
```diff
@@ -53,24 +46,7 @@ interface PageEventBody {
...
-        domain: 'builder.io',
-      },
-    })
-
-    return new NextResponse('Unauthorized', {
-      status: 401,
-    })
-  }
-
+export const POST = withBuilderIoAuthMiddleware(async (request: NextRequest) => {
```
lucasrmp-cb, line 49: You shouldn't remove `withRouteMiddleware` from this route, IIRC it adds some important sentry metadata to the request. You can nest these higher order function calls no problem (or you should be able to if the typings are set correctly)
```suggestion
export const POST = withRouteMiddleware(withBuilderIoAuthMiddleware(async (request: NextRequest) => {
```

lucasrmp-cb (changes requested):

src/app/api/public/builder/events/page/route.ts
olavoparno: does it even make sense to have it around a webhook called by builder.io?

lucasrmp-cb (approved):

PR #1802, opened 2025-01-31
refactor: changes locale to countryCode

### LINKED ISSUE #1780: Update codebase to work with new multi-tenant schema
After updating the schema, we will need to pass the locale property in many different places, which will likely require a lot of code changes. This issue is specific to this work.

twistershark:
closes #1780 

## What changed? Why?

As we don't plan to implement translations anytime soon on our project, I've created this PR to remove references to locales and use country codes instead. The idea is to have URLs with only the country code (e.g., `/ca/` instead of `en-CA`), making it simpler and avoiding the misconception that a French version (`fr-CA`) might exist. This change will also simplify data filtering in Mixpanel, using `US` instead of `en-US`.

With that in mind, this PR updates all references to `locales` and switches to using `countryCode` instead.

There are only a few places where we actually need the `locale`. In these instances, I've created a mapper to map the country code to a specific locale. These places are:

- HTML lang tag
- Currency formatting functions
- Date/Time formatting functions

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/countryCodeRouter/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  const searchParams = request.nextUrl.search
+
+  const segments = pathname.split('/').filter(Boolean)
+  const firstSegment = segments[0] ?? null
+
+  const createUrl = (path: string): URL => new URL(`${path}${searchParams}`, request.url)
+
+  // If path starts with default country code, redirect to path without it
+  if (firstSegment === DEFAULT_SUPPORTED_COUNTRY_CODE) {
+    const newPath = pathname.replace(`/${DEFAULT_SUPPORTED_COUNTRY_CODE}`, '') || '/'
```
olavoparno, line 30: must check for both casings e.g. 'us' and 'US'

olavoparno (dismissed):

lucasrmp-cb (approved):

PR #1833, opened 2025-02-11
feat: investigate and fix middleware cookie creation

### LINKED ISSUE #1828: Investigate and fix middleware countryCode cookie creation
This week, we shipped a new feature to set a countryCode cookie on each request to be used across the application as part of our multi-tenant architectural change approach.

After deploying, we discovered that wallet login stopped working, causing login issues for almost an hour.

We need to investigate and fix this issue to retain the cookie creation logic without creating new bugs.

twistershark:
closes #1828 

## What changed? Why?

This PR re-adds the current page country code cookie creation in our middleware without breaking the sign-in flow for wallets on static pages.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/countryCodeRouter/index.ts
```diff
@@ -1,10 +1,20 @@
...
+  DEFAULT_SUPPORTED_COUNTRY_CODE,
+  SWC_CURRENT_PAGE_COUNTRY_CODE_COOKIE_NAME,
+} from '@/utils/shared/supportedCountries'
 
 // Two lowercase letters (e.g., "us", "uk")
 const COUNTRY_CODE_PATTERN = /^[a-z]{2}$/
 
+function extractCountryCode(pathname: string) {
+  const segments = pathname.split('/').filter(Boolean)
+  const firstSegment = segments[0] ?? null
```
LucasdRossi, line 13: Is this `?? null` really necessary?

olavoparno (approved):

src/utils/server/countryCodeRouter/index.ts
twistershark: Yes! 

I need it to be null because I check if it is null on the router function. 

Not returning null would remove the benefit of this function, because I would need to duplicate the same logic across the save cookie function and the router in this file

LucasdRossi: But you're just using `firstSegment` to do the `COUNTRY_CODE_PATTERN` validation. It doesn't matter if its null or undefined, because either one of them would fail the validation and return null on line 15, right?

twistershark: You are right! I thought you were asking why the function had to return null.

LucasdRossi (approved):

olavoparno (approved):

PR #1842, opened 2025-02-12
feat: creates temporary inngest function to fix uppercase tenant ids

### LINKED ISSUE #1841: Create fix uppercase tenantId inngest function
Because of the time period that we had the application running while the tenantId column migration was running, two new advocates joined SWC, causing some tables to have the tenantId as 'US' instead of 'us'.

We need to create a temporary script to fix those records and then remove the script from the project.

twistershark:
closes #1841

## What changed? Why?

This PR creates a new Inngest script to update all records across our database that have the tenantId as 'US' instead of 'us'.

This script will only be executed in tables that have rows with the wrong tenantId.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/inngest/functions/fixTenantId.ts
+++ b/src/inngest/functions/fixTenantId.ts
@@ -1,0 +1,261 @@
+import { inngest } from '@/inngest/inngest'
+import { prismaClient } from '@/utils/server/prismaClient'
+
+const FIX_TENANT_ID_INNGEST_FUNCTION_ID = 'fix-tenant-id'
+const FIX_TENANT_ID_INNGEST_EVENT_NAME = 'fix.tenant.id'
+
+export interface FixTenantIdInngestEventSchema {
+  name: typeof FIX_TENANT_ID_INNGEST_EVENT_NAME
+  data: {
+    persist: boolean
+  }
+}
+
+export const fixTenantIdWithInngest = inngest.createFunction(
+  {
+    id: FIX_TENANT_ID_INNGEST_FUNCTION_ID,
+    retries: 0,
+  },
+  { event: FIX_TENANT_ID_INNGEST_EVENT_NAME },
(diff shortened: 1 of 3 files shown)


src/inngest/functions/fixTenantId.ts
```diff
@@ -0,0 +1,261 @@
...
+    id: FIX_TENANT_ID_INNGEST_FUNCTION_ID,
+    retries: 0,
+  },
+  { event: FIX_TENANT_ID_INNGEST_EVENT_NAME },
+  async ({ event, step }) => {
+    const { persist } = event.data
+
+    await step.run('update users table', async () => {
+      const usersWithWrongTenantId: Array<{ id: string; tenant_id: string }> =
+        await prismaClient.$queryRaw`
```
twistershark, line 25: I'm using queryRaw here because prisma with PlanetScale doesn't support case sensitive columns by default. We would need to change settings on PlanetScale to support this and that's not something that we want.

With a queryRaw we can collate the column to make the query be case sensitive.

olavoparno:
maybe add a return value of the selects before running it persisted, you know? if persist is false, simply show the records we're changing

olavoparno (approved):

twistershark:
> maybe add a return value of the selects before running it persisted, you know? if persist is false, simply show the records we're changing

For each table that will be updated, I've added a console.table with all records that will be updated. The result in the same as the data on this PR description.

```js
console.log('Users with wrong tenantId')
      console.table(
        usersWithWrongTenantId.map(user => ({
          id: user.id,
          tenantId: user.tenant_id,
          status: 'needs update',
        })),
      )
```

Does this fulfill your need?

LucasdRossi (approved):

PR #1862, opened 2025-02-14
Trigger REFER action creation on signup

### LINKED ISSUE #1861: Referrals - Trigger REFER action creation on signup

eduardopicolo-cb:
closes #1861 

## What changed? Why?

Adds a new post signup step to create a REFER action for the referrer. Leverages NextJS `after` and only triggers it on valid referral signup.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/thirdweb/onLogin.ts
```diff
@@ -476,6 +482,8 @@ export async function onNewLogin(props: NewLoginParams) {
     'Datetime of Last Login': new Date(),
   })
 
+  triggerReferralSteps({ localUser, searchParams, userId: user.id })
```
LucasdRossi, line 485: I think that you should check if this is a referral before calling this function instead of checking it inside of the function, because then it will be easier to understand that we're only taking action if this is a referral, right now if I don't open the function I would think that we're considering every login a referral

src/bin/smokeTests/onNewLogin/12testCaseWithValidReferral.ts
```diff
@@ -0,0 +1,142 @@
...
+  name: 'Test Case with valid referral',
+  parameters: async () => {
+    const { referrerId } = await createTestReferrer()
+    return createTestParams(referrerId)
+  },
+  validateResults: async (data: OnNewLoginResult, issues: string[]) => {
+    verifyBasicUserCreation(data, issues)
+
+    // Wait a bit for the after() function to complete
+    await new Promise(resolve => setTimeout(resolve, 100))
```
LucasdRossi, line 136: We have a utility function called `sleep` its in `@/utils/shared/sleep`

src/actions/actionCreateUserActionReferral.ts
```diff
@@ -0,0 +1,137 @@
...
+import { withSafeParseWithMetadata } from '@/utils/shared/zod'
+
+const logger = getLogger(`actionCreateUserActionReferral`)
+
+const zodUserActionReferralInput = z.object({
+  referralId: z
+    .string()
+    .length(12, 'Referral ID must be exactly 12 characters long')
+    // Zod's nanoid does not allows to specify a custom length for the id
+    .regex(/^[A-Za-z0-9_-]+$/, 'Referral ID contains invalid characters'),
```
LucasdRossi, line 24: You could move this into a separate file in `@/validation`, keep me honest here, but I think that this will be used multiple times to validate if the id is correct, so we could just have a `zodReferralId`

LucasdRossi (approved):

PR #1864, opened 2025-02-18
Referrals ranking cron job

### LINKED ISSUE #1837: Referrals - Inngest cron job

eduardopicolo-cb:
closes #1837 

## What changed? Why?

- Added Inngest cron job to query the district data and upsert it in Redis;
- Compression is not needed, as the final size is about 45KB; 
- Added an index for address queries;

image

image

## PlanetScale deploy request

https://app.planetscale.com/stand-with-crypto/swc-web/deploy-requests/98

## Notes to reviewers

🗒️ Referrals count looks way higher in testing due to the mocking script generating a lot more REFER actions than users;

- Added DC state to our mock data with a probability of 5%, this is because faker does not consider DC as an State 😞 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -51,7 +51,6 @@ model Address {
   userActionVotingInformationResearched UserActionVotingInformationResearched[]
 
   @@index([administrativeAreaLevel1, countryCode])
-  @@index([countryCode, usCongressionalDistrict, administrativeAreaLevel1])
   @@map("address")
 }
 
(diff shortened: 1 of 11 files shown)


prisma/schema.prisma
```diff
@@ -51,6 +51,7 @@ model Address {
   userActionVotingInformationResearched UserActionVotingInformationResearched[]
 
   @@index([administrativeAreaLevel1, countryCode])
+  @@index([countryCode, usCongressionalDistrict, administrativeAreaLevel1])
```
LucasdRossi, line 54: Do you have any evidence showing that this new index is necessary?

eduardopicolo-cb: After running the query on a prod replica with EXPLAIN ANALYSE the index is actually slowing down, so im removing it

LucasdRossi (approved):

PR #1875, opened 2025-02-19
feat: save tenantId across the codebase

### LINKED ISSUE #1783: Update actions/scripts to save tenantId information
We can handle this by passing the tenantId prop where we call the scripts and actions, or we can use the `SWC_LOCALE` cookie to get the current country code value, enabling us to create data specific to each tenant.

Here is a list of possible places that need to be updated:

- src/actions/actionCreateUserActionEmailCongressperson.ts
- src/actions/actionCreateUserActionEmailDebate.ts
- src/actions/actionCreateUserActionLiveEvent.ts
- src/actions/actionCreateUserActionNFTMint.ts
- src/actions/actionCreateUserActionRsvpEvent.ts
- src/actions/actionCreateUserActionTweet.ts
- src/actions/actionCreateUserActionTweetAtPerson.ts
- src/actions/actionUpdateUserProfile.ts
- src/actions/actionCreateUserActionCallCongressperson/index.ts
- src/app/api/internal/in-app-donations/route.ts
- src/bin/localDevelopment/simulateLiveActivityOnSite.ts
- src/bin/seed/seedLocalDb.ts
- src/bin/seed/setUpLocalUserToHaveEmbeddedEmailAddress.ts
- src/bin/seed/setUpLocalUserToHaveLegacyCryptoAddress.ts
- src/bin/smokeTests/nft/airdropNFT.ts
- src/bin/smokeTests/onNewLogin/2testCaseUserPreviouslySignedUpOnCoinbaseAndThenLoggedInWithEmbeddedWalletWithSameEmail.ts
- src/bin/smokeTests/onNewLogin/3testCaseUserHasLegacyMigrationCryptoAddressAndEmailAndLogsOnViaSessionIdWithNewCryptoAddress.ts
- src/bin/smokeTests/onNewLogin/4testCaseUserHasLegacyMigrationCryptoAddressAndEmailAndLogsOnViaSameCryptoAddress.ts
- src/bin/smokeTests/onNewLogin/5testCaseUserHasLegacyMigrationCryptoAndEmailAddressAndLogsOnViaSameCryptoAndEmailAddressButAlreadyHasCreatedUsers.ts
- src/bin/smokeTests/onNewLogin/6testCaseUserHasSameEmailAddressButOtherUserHasAlreadyCreatedAccount.ts
- src/bin/smokeTests/onNewLogin/7testCaseMultipleUsersWithSameLegacyCryptoAddress.ts
- src/bin/smokeTests/onNewLogin/8testCaseMultipleUsersWithSameLegacyEmail.ts
- src/bin/smokeTests/onNewLogin/11testCaseWithPhoneSignUpWithExistingAccount.ts
- src/inngest/functions/initialSignupUserCommunicationJourney/initialSignupUserCommunicationJourney.tsx
- src/utils/server/coinbaseCommerce/storePaymentRequest.ts
- src/utils/server/externalOptIn/handleExternalUserActionOptIn.ts
- src/utils/server/sms/communicationJourney.ts
- src/utils/server/thirdweb/onLogin.ts
- src/actions/actionUpdateUserHasOptedInSMS.ts

twistershark:
closes #1783 

## What changed? Why?

This PR updates all required server functions to pass the correct user tenant ID. With this change, we will be able to remove the default property from the tenantId column on some tables in our database.

## Notes to reviewers

We'll want to bug bash this PR before merging

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/mocks/models/mockUserActionCall.ts
```diff
@@ -8,6 +8,7 @@ export function mockCreateUserActionCallInput() {
   return {
     recipientPhoneNumber: fakerFields.phoneNumber(),
     recipientDtsiSlug: fakerFields.dtsiSlug(),
+    tenantId: faker.helpers.arrayElement(Object.values(ORDERED_SUPPORTED_COUNTRIES)),
```
olavoparno, line 11: we prob don't need the Object.values here since ORDERED_SUPPORTED_COUNTRIES is already an array

src/utils/server/coinbaseCommerce/storePaymentRequest.ts
```diff
@@ -205,12 +206,17 @@ async function createNewUser(payment: CoinbaseCommercePayment) {
       acquisitionSource: 'coinbase-commerce-webhook',
       acquisitionMedium: '',
       acquisitionCampaign: '',
+      // Defaulting to US here as we don't expect to support donations in other countries.
+      // If we do, then we need to find a way to get the user's country code and update this accordingly.
+      tenantId: DEFAULT_SUPPORTED_COUNTRY_CODE,
```
olavoparno, line 211: I think we should indeed hardcode this for the very reason you mentioned but we could pass in from the route or action calling this function, couldn't we?

src/utils/server/getTenantId.ts
```diff
@@ -0,0 +1,27 @@
...
+      throw new Error('Tenant ID cookie not found')
+    }
+
+    if (!COUNTRY_CODE_REGEX_PATTERN.test(tenantId)) {
+      throw new Error('Invalid Tenant ID cookie')
+    }
+
+    return tenantId
+  } catch (error) {
+    Sentry.captureException(error)
```
olavoparno, line 24: we should probably add more metadata here. We're getting the error stack in full, right, but maybe add current cookies or the very tenantId to it?

olavoparno (dismissed):

src/utils/server/getTenantId.ts
twistershark: Passing all cookies is a security problem as we would be sending sensitive user information to Sentry(auth token, for example).

Agree that we could add more metadata, but not sure how. I've updated the condition that checks if the parsed cookie is a valid country code to also pass the tenantId so that we have more information on what value the cookie had that caused the comparison to fail.


Other than that, I don't think there is much else that I could do.

For example: `await cookies()` is inside the try/catch block. If I moved it outside of the block to pass some cookies as metadata, we wouldn't catch errors that it could throw because the `await cookies()` call failed.

We also don't have access to where this function is being called. But this shouldn't be an issue because Sentry already adds information like this when an error is captured.

olavoparno: we can always feed more info inside a try without the need to move the function call outside of the try by adding more context, tags or extra to sentry using its setters e.g. setExtra, setContext, etc but we may already have the info for tentantId on Sentry errors through the cookie SWC_CURRENT_PAGE_COUNTRY_CODE. Thanks man

lucasrmp-cb: Agree on not adding all cookies, but tanking the `await cookies()` call out of the try/catch block and then adding the country code cookie value as additional data would be nice, since it not existing or being invalid are the only failure condition for the function

src/validation/fields/zodAddress.ts
```diff
@@ -13,4 +13,5 @@ export const zodAddress = object({
   postalCodeSuffix: string(),
   countryCode: string().length(2),
   usCongressionalDistrict: string().optional(),
+  tenantId: string().length(2).optional(),
```
lucasrmp-cb, line 16: We can have a more strict validation for this string's value
https://zod.dev/?id=zod-enums

olavoparno (dismissed):

lucasrmp-cb (dismissed):

src/validation/fields/zodAddress.ts
twistershark: Updated

olavoparno (approved):

lucasrmp-cb (approved):

PR #1891, opened 2025-02-25
feat: save country code information in mixpanel and sentry

### LINKED ISSUE #1784: Save locale metadata in Sentry and Mixpanel
To have specific metrics for each locale, we will leverage the SWC_LOCALE cookie that is [created in our middleware logic ](https://github.com/Stand-With-Crypto/swc-web/issues/1781) to update the corresponding locale as metadata for Mixpanel and Sentry. 

A potential problem we might encounter is trying to update the metadata before Mixpanel and Sentry are initialized, as all this logic will be executed inside the middleware. If that is the case, we will try to set the metadata during Mixpanel and Sentry initialization, leveraging the SWC_LOCALE cookie that was created as described in the previous section.

twistershark:
closes #1784 

## What changed? Why?

This PR saves the country code cookie on Mixpanel events and sentry to help track and filter events and errors by country code. 

Note that I'm already using countryCode naming instead of tenantId because we've decided to pivot from the multi-tenant architecture. All other necessary changes to rename tenantId to countryCode will be done afterwards.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/server/thirdweb/onLogin.ts
+++ b/src/utils/server/thirdweb/onLogin.ts
@@ -362,7 +362,6 @@ export async function onNewLogin(props: NewLoginParams) {
       localUser,
       hasSignedInWithEmail,
       sessionId: await props.getUserSessionId(),
-      tenantId,
     }).catch(error => {
       log(
         `createUser: error creating user\n ${JSON.stringify(
@@ -598,12 +597,10 @@ async function createUser({
   localUser,
   hasSignedInWithEmail,
   sessionId,
-  tenantId,
 }: {
   localUser: ServerLocalUser | null
   hasSignedInWithEmail: boolean
   sessionId: string | null
-  tenantId: string
 }) {
   return prismaClient.user.create({
     include: {
@@ -621,7 +618,6 @@ async function createUser({
       userSessions: {
         create: { id: sessionId ?? undefined },
       },
-      tenantId,
       ...mapLocalUserToUserDatabaseFields(localUser),
     },
   })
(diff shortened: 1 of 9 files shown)


src/utils/web/clientLocalUser.ts
```diff
@@ -71,7 +78,9 @@ export const getLocalUser = (): LocalUser => {
     }
     return localUser
   }
-  const currentSession: CurrentSessionLocalUser = getDefaultCurrentSessionLocalUser()
+
+  const currentSession: CurrentSessionLocalUser = getDefaultCurrentSessionLocalUser(countryCode)
```
LucasdRossi, line 82: The return of the function is already typed `CurrentSessionLocalUser`, so you don't need to type it again

src/utils/server/thirdweb/onLogin.ts
```diff
@@ -598,12 +597,10 @@ async function createUser({
   localUser,
   hasSignedInWithEmail,
   sessionId,
-  tenantId,
```
LucasdRossi, line 601: Why are we removing this?

twistershark: Because if is not needed anymore. TenantId will be deprecated

src/utils/web/clientLocalUser.ts
twistershark: Updated

lucasrmp-cb (commented):
we're missing server analytics

lucasrmp-cb (approved):

PR #1938, opened 2025-02-26
fix: add error handling to queryDTSIBillDetails

LucasdRossi:
fixes [PROD-SWC-WEB-5TD](https://stand-with-crypto.sentry.io/issues/6302930936/events/527f367097264a86a95f3c627f6abab4/)

## What changed? Why?

Added a catch to queryDTSIBillDetails so the page redirects to 404 instead of crashing when accessing a invalid bill

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/data/dtsi/queries/queryDTSIBillDetails/index.ts
+++ b/src/data/dtsi/queries/queryDTSIBillDetails/index.ts
@@ -1,3 +1,5 @@
+import * as Sentry from '@sentry/nextjs'
+
 import { fetchDTSI } from '@/data/dtsi/fetchDTSI'
 import { DTSI_BillDetailsQuery, DTSI_BillDetailsQueryVariables } from '@/data/dtsi/generated'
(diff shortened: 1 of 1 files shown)


src/data/dtsi/queries/queryDTSIBillDetails/index.ts
```diff
@@ -4,13 +4,9 @@ import { dtsiBillDetailsQueryString } from '@/data/dtsi/queries/queryDTSIBillDet
...
-  )
-
-  return results.bill
-}
+export const queryDTSIBillDetails = (id: string) =>
+  fetchDTSI<DTSI_BillDetailsQuery, DTSI_BillDetailsQueryVariables>(dtsiBillDetailsQueryString, {
+    id,
+  })
+    .then(res => res.bill)
+    .catch(() => null)
```
eduardopicolo-cb, line 12: Can we add Sentry here and add the bill id

eduardopicolo-cb (changes requested):

eduardopicolo-cb (approved):

Issue #1896, opened 2025-02-25
Update address field on user profile update to call the new server action to update the user's country code


lucasrmp-cb:
- update user's country code based on address update on profile page (that address country being supported by SWC)
- add a disclaimer before updating user's country code based on new address country code (if supported by SWC)

TODO: review copy and UX for that scenario

olavoparno:
current suggestion for a disclaimer when user is about to change their countryCode:

![Image](https://github.com/user-attachments/assets/0ad4baec-98e3-480b-ac4c-9b2ce0b8d96a)

PR #1941, opened 2025-02-26
Feat/update user profile country code

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1896

fixes 

## What changed? Why?

image

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/updateUserProfileForm/step1/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { ORDERED_SUPPORTED_COUNTRIES } from '@/utils/shared/supportedCountries'
+
+export const DEFAULT_DISCLAIMER =
+  'You are about to change your viewing country based on the address you have chosen'
+
+export const DISCLAIMERS_BY_COUNTRY_CODE: Record<
+  (typeof ORDERED_SUPPORTED_COUNTRIES)[number],
```
lucasrmp-cb, line 7: No need for this, you can just use `SupportedCountryCodes`
```suggestion
  SupportedCountryCodes,
```

src/components/app/updateUserProfileForm/step1/index.tsx
```diff
@@ -81,39 +85,78 @@ export function UpdateUserProfileForm({
...
+      const TEMPORARY_ALLOWED_COUNTRY_CODE = 'gb' // TODO(@olavoparno): Remove this once we start supporting UK inside zodSupportedCountryCode
+
+      const isCountryCodeSupported =
+        success || addressCountryCode === TEMPORARY_ALLOWED_COUNTRY_CODE
+
+      if (isCountryCodeSupported && validatedAddressCountryCode !== userCountryCode) {
+        setShouldShowCountryCodeDisclaimer(true)
+      }
+    }
+  }, [resolvedAddress, user.countryCode])
```
lucasrmp-cb, line 133: Let's create an address field to contain this logic

src/components/app/updateUserProfileForm/step1/index.tsx
```diff
@@ -81,39 +85,78 @@ export function UpdateUserProfileForm({
...
+
+  useEffect(() => {
+    if (resolvedAddress) {
+      const addressCountryCode = resolvedAddress.countryCode.toLowerCase()
+      const userCountryCode = user.countryCode.toLowerCase()
+
+      const { success, data: validatedAddressCountryCode } =
+        zodSupportedCountryCode.safeParse(addressCountryCode)
+
+      const TEMPORARY_ALLOWED_COUNTRY_CODE = 'gb' // TODO(@olavoparno): Remove this once we start supporting UK inside zodSupportedCountryCode
```
lucasrmp-cb, line 124: We're never supporting "uk" as a country, gb is the right one for England

src/validation/fields/zodSupportedCountryCode.ts
```diff
@@ -10,6 +10,4 @@ export const zodSupportedCountryCode = string()
     if (lowerCaseCountryCode === 'gb') return 'uk'
     return lowerCaseCountryCode
   })
```
lucasrmp-cb, line 12: According to @twistershark this shouldn't be here, let's remove this check

src/components/app/updateUserProfileForm/step1/index.tsx
lucasrmp-cb: Let's remove this check

lucasrmp-cb (changes requested):

lucasrmp-cb (changes requested):

src/actions/actionUpdateUserCountryCode.ts
```diff
@@ -92,6 +100,21 @@ async function actionUpdateUserCountryCodeWithoutMiddleware(
     },
   })
 
+  const localUser = await parseLocalUserFromCookies()
+
+  waitUntil(
```
lucasrmp-cb, line 105: One more thing, you're changing the user's properties but not registering the event of changing the country code, let's add a server analytics call (not people analytics) and create an event of changing country code

PR #1945, opened 2025-02-27
refactor: limits use of `getCountryCodeOnClient` function to analytics only

twistershark:
## What changed? Why?

This PR refactors the use of the getCountryCodeOnClient function to be only in analytics environments where having the default 'not-set' property is not a problem.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/web/getCountryCodeOnClient.ts
+++ b/src/utils/web/getCountryCodeOnClient.ts
@@ -1,11 +1,7 @@
-import * as Sentry from '@sentry/nextjs'
 import Cookies from 'js-cookie'
 
 import { isBrowser } from '@/utils/shared/executionEnvironment'
-import {
-  COUNTRY_CODE_REGEX_PATTERN,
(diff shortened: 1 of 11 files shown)


src/utils/web/getCountryCodeOnClient.ts
```diff
@@ -16,11 +12,5 @@ export function getCountryCodeOnClient() {
     return 'not-set'
   }
 
-  if (!COUNTRY_CODE_REGEX_PATTERN.test(countryCode)) {
```
twistershark, line 19: We don't need this validation anymore as this function is only used in analytics context and having this validation here only results in blocks and could create unexpected bugs.

olavoparno (approved):

olavoparno (approved):

lucasrmp-cb (approved):

PR #1957, opened 2025-02-28
Feat/update user cookie on login

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1899

fixes 

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/utils/server/thirdweb/onLogin.ts
+++ b/src/utils/server/thirdweb/onLogin.ts
@@ -137,6 +137,7 @@ export async function login(payload: VerifyLoginPayloadParams) {
         httpOnly: false,
         sameSite: 'lax',
         secure: true,
+        maxAge: 24 * 60 * 60 * 1000, // 24 hours
       })
     }
 
--- a/src/utils/server/thirdweb/onLogin.ts
+++ b/src/utils/server/thirdweb/onLogin.ts
@@ -16,7 +16,7 @@ import {
 } from '@prisma/client'
 import * as Sentry from '@sentry/nextjs'
 import { waitUntil } from '@vercel/functions'
-import { compact, groupBy, values } from 'lodash-es'
+import { compact, groupBy } from 'lodash-es'
 import { cookies } from 'next/headers'
 import { VerifyLoginPayloadParams } from 'thirdweb/auth'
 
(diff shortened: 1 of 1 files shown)


src/utils/server/thirdweb/onLogin.ts
```diff
@@ -114,29 +114,39 @@ export async function login(payload: VerifyLoginPayloadParams) {
...
       parsedUserCountryCodeCookie?.countryCode.toLowerCase() !==
         currentUserCountryCode.toLowerCase()
     ) {
-      currentCookies.set(
-        USER_COUNTRY_CODE_COOKIE_NAME,
-        JSON.stringify({ countryCode: currentUserCountryCode.toLowerCase(), bypassed: false }),
-        { sameSite: 'lax' },
-      )
+      log(`setting user country code cookie to ${currentUserCountryCode.toLowerCase()}`)
+      currentCookies.set({
```
twistershark, line 131: Please, add a `maxAge: 24 * 60 * 60 * 1000, // 24 hours` so that we don't need to implement new logic to handle deleting the cookie based on different conditions

src/utils/server/thirdweb/onLogin.ts
```diff
@@ -114,29 +114,40 @@ export async function login(payload: VerifyLoginPayloadParams) {
...
-      context: { userId: existingVerifiedUser.id, address },
-    })
-
     const currentUserCountryCode = existingVerifiedUser.countryCode
 
     const userCountryCodeCookie = currentCookies.get(USER_COUNTRY_CODE_COOKIE_NAME)?.value
     const parsedUserCountryCodeCookie = parseUserCountryCodeCookie(userCountryCodeCookie)
 
-    currentCookies.set(THIRDWEB_AUTH_TOKEN_COOKIE_PREFIX, jwt)
+    log(`currentUserCountryCode: ${currentUserCountryCode}`)
```
twistershark, line 122: I would suggest keeping only the log `setting user country code cookie to` and if we want to keep it, I would suggest updating it to: `setting user country code cookie from ${parsedUserCountryCodeCookie?.countryCode} to ${currentUserCountryCode.toLowerCase()}`

twistershark (approved):

PR #1958, opened 2025-02-28
Feature/poll

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-internal/issues/342
 
fixes 

## How has it been tested?

- [X] Locally
- [X] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


bin/vercel_deploy.sh
```diff
@@ -20,7 +20,7 @@ checkDbSchema() {
 }
 
 # removed temporarily until we have a better way to check the DB schema
-# checkDbSchema
+checkDbSchema
```
twistershark, line 23: Just a suggestion: IMO, it's better to do changes like this in different PRs with each one having a specific context. But you don't need to change anything for this PR.

prisma/schema.prisma
```diff
@@ -362,6 +364,29 @@ model UserActionEmailRecipient {
   @@map("user_action_email_recipient")
 }
 
+model UserActionPoll {
+  id                    String                 @id @default(uuid())
+  userAction            UserAction             @relation(fields: [id], references: [id], onDelete: Cascade)
+  tenantId              String                 @default("us") @map("tenant_id")
```
twistershark, line 370: We don't need this anymore. We now use the country_code column and it is only in the UserAction and Users table

prisma/schema.prisma
```diff
@@ -362,6 +364,29 @@ model UserActionEmailRecipient {
...
+
+model UserActionPollAnswer {
+  id                     String         @id @default(uuid())
+  userActionPollId       String         @map("user_action_poll_id")
+  userActionPoll         UserActionPoll @relation(fields: [userActionPollId], references: [id], onDelete: Cascade)
+  // denormalized from userAction's campaignName to allow for easier querying
+  userActionCampaignName String         @map("user_action_campaign_name")
+  answer                 String         @map("answer")
+  isOtherAnswer          Boolean        @map("is_other_answer")
+  tenantId               String         @default("us") @map("tenant_id")
```
twistershark, line 384: The same as above

src/hooks/useSession.ts
```diff
@@ -22,8 +22,10 @@ export function useSession() {
 
   const isLoggedIn = thirdwebSession.isLoggedIn || hasOptInUserAction
   const isLoggedInThirdweb = thirdwebSession.isLoggedIn
+
   return {
     isLoading: !isLoggedIn && isLoading,
```
twistershark, line 27: Nit picking: I would rename this isLoading to `isLoadingNonAuthenticated` and would set the new property instead of `isUserProfileLoading` to `isLoading`

src/components/app/authentication/loginDialogWrapper.tsx
```diff
@@ -209,7 +220,7 @@ export function UnauthenticatedSection({
         className="max-w-l w-full px-0 md:px-0"
       >
         <GeoGate
```
twistershark, line 222: I would just remove the GeoGate here.

prisma/schema.prisma
olavoparno: thanks, it's been some time since the last changes 🤣

src/components/app/pagePolls/pollLegend.tsx
```diff
@@ -0,0 +1,30 @@
...
+    )
+  }
+
+  return (
+    <span className="text-sm text-gray-500">
+      {hasEnded
+        ? `Ended on ${format(new Date(endDate), 'MMM d, yyyy')}`
+        : `Ends in ${endsIn} ${dayOrDays}`}
+    </span>
+  )
```
twistershark, line 29:
```suggestion
  const endDate = new Date(endDate)
  const endsIn = differenceInDays(endDate, new Date())
  const hasEnded = isPast(endDate)
  const hasEndedToday = endsIn === 0
  const dayOrDays = endsIn === 1 ? 'day' : 'days'

  if (hasEndedToday) {
    return (
      <span className="text-sm text-gray-500">{`${isInactivePoll ? 'Ended' : 'Ends'} today`}</span>
    )
  }

  return (
    <span className="text-sm text-gray-500">
      {hasEnded
        ? `Ended on ${format(endDate, 'MMM d, yyyy')}`
        : `Ends in ${endsIn} ${dayOrDays}`}
    </span>
  )
```

src/utils/shared/getSWCPolls.ts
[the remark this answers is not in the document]
olavoparno: I'd have to move everything else not related to Polls as well but I agree with you. I will move them accordingly

src/components/app/pagePolls/geoGatedPollsContent.tsx
```diff
@@ -0,0 +1,60 @@
...
+  activePoll: SWCPoll
+  shouldShowResults: boolean
+  handleShowResults: () => void
+  handleVoteAgain: () => void
+  isLoading: boolean
+  pollsResults: Record<string, PollResultsDataResponse>
+  userPolls: PollsVotesFromUserResponse | undefined
+}
+
+export function GeoGateContent({
```
twistershark, line 19: I would rename this function to make it clear that this is related to polls. We prevent someone thinking that there is a generic GeoGateContent component that can be used in different contexts

src/hooks/useSession.ts
olavoparno: I have created a new property `isUserProfileLoading` because we only relied on the `isLoading` which could lead to incorrect state. The property `isUserProfileLoading` just shows if we're still fetching the user profile.

twistershark (approved):

src/components/app/authentication/loginDialogWrapper.tsx
twistershark: Because when a new user signs up, we save the country code related to the page they signed up, not the IP country code. With that, we don't need to worry about having the GeoGate because it won't have any result and it is currently only adding unnecessary code. You don't need to remove it in this PR as it is not directly related, but as you had to pass a new prop to it, I thought it could be a good opportunity to remove it

Check this PR to see how the new country code logic works: https://github.com/Stand-With-Crypto/swc-web/pull/2012

lucasrmp-cb (approved):

twistershark (approved):

PR #1960, opened 2025-03-03
feat: add international websites

### LINKED ISSUE #1913: SWC International - Build intl websites
This is the epic for all the feature work we need to deliver the international websites

lucasrmp-cb:
closes #1913 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/app/[countryCode]/layout.tsx
```diff
@@ -29,33 +26,18 @@ export { viewport } from '@/utils/server/metadataUtils'
...
 export default async function Layout({
   children,
   params,
 }: PageProps & { children: React.ReactNode }) {
   const { countryCode } = await params
 
-  if (!ORDERED_SUPPORTED_COUNTRIES.includes(countryCode)) {
+  if (countryCode !== DEFAULT_SUPPORTED_COUNTRY_CODE) {
     notFound()
   }
```
Comment (olavoparno) [inline src/app/[countryCode]/layout.tsx:42]: prob need normalization to lowercase on these cases because countryCode here is untreated

src/app/au/error.tsx
```diff
@@ -0,0 +1,20 @@
...
+  reset,
+}: {
+  error: Error & { digest?: string }
+  reset: () => void
+}) {
+  useHandlePageError({
+    domain: 'rootErrorPage',
+    humanReadablePageName: 'Root',
+    error,
+  })
```
olavoparno, line 17: we should prob append to the domain name a differentiation or add countryCode to tags inside useHandlePageError

src/components/app/layout/layout.tsx
olavoparno, line 1: the file naming could be improved but a note here, should we add html (root layout stuff) abstracted in files other than `layout.tsx`? We could put html and body tags inside each country specific layout file and the rest could be in this file

src/app/[countryCode]/layout.tsx
Comment (lucasrmp-cb) [reply src/app/[countryCode]/layout.tsx:42]: It didn't have it before, I'm not changing behaviors

src/app/au/error.tsx
lucasrmp-cb: It's already being done through the sentry integration, we don't need to change the calls

src/components/app/layout/layout.tsx
lucasrmp-cb: You'll have an error if you try, and for me the file name having the same name as the app directory convention makes sense.
And about having html and body specific to layouts, they have specific props that are either default (font classNames) or dynamic based on country code (language). Since the only use case this is supposed to provide is a page layout I'd say this makes sense. I'd prefer waiting to face a limitation before making it more (or differently) abstracted

olavoparno (dismissed):

PR #1978, opened 2025-03-06
Allow custom utm_medium query param in the /join vanity url

### LINKED ISSUE #1977: Preserve utm_medium parameter in /join redirects

eduardopicolo-cb:
closes #1977 

fixes 

## What changed? Why?

Previously, all redirects from `/join/:referralId` would set a fixed `utm_medium=referral` value, causing any original utm_medium values to be lost. This change ensures accurate attribution tracking by:
1.  Capturing the original utm_medium value when present
2.  Maintaining the default referral value when no utm_medium is provided

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -294,8 +294,28 @@ const nextConfig: NextConfig = {
       // vanity urls
       {
         source: '/join/:referralId',
+        destination:
+          '/action/sign-up?utm_campaign=:referralId&utm_source=swc&utm_medium=:utm_medium',
+        permanent: false,
+        has: [
+          {
+            type: 'query',
+            key: 'utm_medium',
+            value: '(?<utm_medium>.*)',
+          },
+        ],
+      },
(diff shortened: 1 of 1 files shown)


next.config.ts
```diff
@@ -292,10 +292,30 @@ const nextConfig: NextConfig = {
...
+      {
+        source: '/join/:referralId',
+        destination:
+          '/action/sign-up?utm_campaign=:referralId&utm_source=swc&utm_medium=:utm_medium',
+        permanent: false,
+        has: [
+          {
+            type: 'query',
+            key: 'utm_medium',
+            value: '(?<utm_medium>.*)',
```
LucasdRossi, line 304: Does this work if utm_medium is an empty string?

LucasdRossi (approved):

twistershark (approved):

Issue #1959, opened 2025-03-03
Create a custom header and footer for the intl pages


lucasrmp-cb:
This needs to abstract header and footer into reusable components and replace the generic ones inside the countries layouts

PR #1990, opened 2025-03-10
Feat/add custom navbar n footer

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/1959

fixes 

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/actions/actionUpdateUserCountryCode.ts
+++ b/src/actions/actionUpdateUserCountryCode.ts
@@ -2,14 +2,18 @@
 import 'server-only'
 
 import * as Sentry from '@sentry/nextjs'
+import { waitUntil } from 'node_modules/@vercel/functions/wait-until'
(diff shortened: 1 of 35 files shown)


twistershark (approved):
Approving for now, but we need @lucasrmp-cb feedback on this when he is back because this [...catchAll] seems odd

olavoparno:
> Approving for now, but we need @lucasrmp-cb feedback on this when he is back because this [...catchAll] seems odd

Thanks @twistershark just for ref, I've added this to temporarily solve the issue of nested not-found not being handled correctly, i.e., if we navigate to /ca/non-existent-path, without [...catchAll] we would ALWAYS render the root not-found.tsx file.

PR #2000, opened 2025-03-13
refactor: simplify page content retrieval by removing countryCode par…

olavoparno:
…ameter

* Updated multiple page components to remove the countryCode parameter from getPageContent and getPageDetails functions.
* Adjusted RenderBuilderContent component to eliminate countryCode prop.
* Introduced new layout and client logic files for GB country code, establishing a foundation for country-specific pages.

closes 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/defaultCountryCodeLayout.tsx
```diff
@@ -9,12 +12,13 @@ import { CookieConsent } from '@/components/app/cookieConsent'
...
 import { FullHeight } from '@/components/ui/fullHeight'
-import { DEFAULT_SUPPORTED_COUNTRY_CODE } from '@/utils/shared/supportedCountries'
+import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
 import { fontClassName } from '@/utils/web/fonts'
 
-export const dynamic = 'error'
-
 export function DefaultCountryCodeLayout({ children }: { children: React.ReactNode }) {
+  const pathname = usePathname()
+  const countryCode = pathname?.split('/')[1] as SupportedCountryCodes
```
LucasdRossi, line 20: @twistershark is this the right way of getting the country code on client?

src/app/not-found.tsx
```diff
@@ -3,8 +3,6 @@ import '@/globals.css'
 import { DefaultCountryCodeLayout } from '@/components/app/defaultCountryCodeLayout'
 import { NotFoundPagesContent } from '@/components/app/notFoundPagesContent'
 
-export const dynamic = 'error'
```
LucasdRossi, line 6: Why did you remove this?

src/app/(builder)/[countryCode]/layout.tsx
```diff
@@ -1,26 +1,13 @@
...
 
-import { TopLevelBuilderClientLogic } from '@/app/(builder)/[countryCode]/topLevelBuilderClientLogic'
-import { TopLevelClientLogic } from '@/app/[countryCode]/topLevelClientLogic'
-import { CookieConsent } from '@/components/app/cookieConsent'
-import { GoogleTagManager } from '@/components/app/googleTagManager'
-import { NavBarGlobalBanner } from '@/components/app/navbarGlobalBanner'
-import { OverrideGlobalLocalStorage } from '@/components/app/overrideGlobalLocalStorage'
-import { FullHeight } from '@/components/ui/fullHeight'
-import { Toaster } from '@/components/ui/sonner'
+import { BuilderLayout } from '@/components/app/builder/builderLayout'
```
Comment (LucasdRossi) [inline src/app/(builder)/[countryCode]/layout.tsx:5]: Is there a reason to move this to a separate file? If we're only using it here I don't see a reason to move it

Comment (olavoparno) [reply src/app/(builder)/[countryCode]/layout.tsx]: it was because of that other idea of having scoped country folders. Forgot to remove, thanks

src/app/not-found.tsx
olavoparno: thanks. due to use of dynamic APIs I had to remove it but it is no longer the case

twistershark (approved):

PR #2024, opened 2025-03-19
Setup action configuration

### LINKED ISSUE #1923: Setup action configuration
The objective here is NOT to build user actions for other countries, all we wanna do is figure out and implement how the hard coded campaign information for different countries will look like.
As a result we want the boilerplate done so other engs can just work on implementing actions

eduardopicolo-cb:
closes #1923 

## What changed? Why?

- Defines the boilerplate to build country-specific user actions
- Updated docs

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


lucasrmp-cb (commented):
I need to have a talk about the overall idea here before approving, let's talk on our 1:1 tomorrow

twistershark (approved):
Are we going to do a bugbash on this?

PR #2054, opened 2025-03-25
Backfill users script

### LINKED ISSUE #1914: Create backfill script
We need to do this for each intl website

We need to get the current users dump from the intl team

The deliverable on this task should be an inngest function where we'd pass a country code and a csv and it inserts all users correctly on the database.
If the spreadsheet is necessary to start work we can table this until we have the dump from the intl team

The backfill has to create the OPT_IN action for all inserted users and account for users who are already signed up to SWC US

- [ ] UK
- [ ] CA
- [ ] AU

eduardopicolo-cb:
closes #1914 

fixes 

## What changed? Why?

- Added `zodPhoneNumberWithCountryCode` to parse phone numbers from different countries

Example user backfilled from csv: Name, email, OPT_IN action and sessionId successfully added to the db
image

## Notes to reviewers

- 🗒️  ~~When intl team provides the CSVs, the`createUserDataSchema` must be updated to match the columns, or edit the CSV itself~~
- Intl team still has to provide an CSV for AU, current implementation assumes it will have the same columns as UK & CA

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/inngest/functions/backfillIntlUsers/logic.ts
+++ b/src/inngest/functions/backfillIntlUsers/logic.ts
@@ -4,6 +4,7 @@ import {
   User,
   UserActionOptInType,
   UserActionType,
+  UserEmailAddressSource,
   UserInformationVisibility,
 } from '@prisma/client'
 import { NonRetriableError } from 'inngest'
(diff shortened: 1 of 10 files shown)


lucasrmp-cb (changes requested):
Let's leave this on hold until we hear back from the intl team with the user dump

src/inngest/functions/backfillIntlUsers/logic.ts
```diff
@@ -0,0 +1,251 @@
...
+      const result = await handleExistingUser({
+        existingUser: existingUser.user,
+        countryCode,
+        persist,
+      })
+
+      return {
+        ...result,
+        email: existingUser.emailAddress,
+      }
```
lucasrmp-cb, line 143: Let's skip on this case, if the user created an account in `standwithcrypto.org` using it's email address we shouldn't just change it to another country

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

PR #2060, opened 2025-03-26
feat: creates AU, CA and GB politicians page

### LINKED ISSUE #2035: Politicians page and home section

### LINKED ISSUE #2053: Create politicians page UI components

twistershark:
closes #2053 

## What changed? Why?

This PR creates `politicians` page for each country.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


olavoparno (approved):

olavoparno (approved):

olavoparno (approved):

src/app/api/public/dtsi/all-people/[countryCode]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+) {
+  const params = await props.params
+  const { countryCode } = zodParams.parse(params)
+
+  if (!countryCode) {
+    return NextResponse.json(
+      { error: 'Country code is required in all-people dtsi route' },
+      { status: 400 },
+    )
+  }
```
Comment (olavoparno) [inline src/app/api/public/dtsi/all-people/[countryCode]/route.ts:29]: maybe change to safeParse or remove the if (!countryCode) entirely?

src/app/au/politicians/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const dynamic = 'error'
+
+type Props = PageProps
+
+export async function generateMetadata(_props: Props): Promise<Metadata> {
+  return generateMetadataDetails({
+    title: PAGE_POLITICIANS_TITLE,
+    description: PAGE_POLITICIANS_DESCRIPTION,
+  })
+}
```
olavoparno, line 24: np: could just be the variable metadataDetails instead of a function

src/app/ca/politicians/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const dynamic = 'error'
+
+type Props = PageProps
+
+export async function generateMetadata(_props: Props): Promise<Metadata> {
+  return generateMetadataDetails({
+    title: PAGE_POLITICIANS_TITLE,
+    description: PAGE_POLITICIANS_DESCRIPTION,
+  })
+}
```
olavoparno, line 24: same here for the static metadata instead of the function

src/app/gb/politicians/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const dynamic = 'error'
+
+type Props = PageProps
+
+export async function generateMetadata(_props: Props): Promise<Metadata> {
+  return generateMetadataDetails({
+    title: PAGE_POLITICIANS_TITLE,
+    description: PAGE_POLITICIANS_DESCRIPTION,
+  })
+}
```
olavoparno, line 24: same here for the static metadata instead of the function

src/app/au/politicians/person/[dtsiSlug]/opengraph-image.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const runtime = 'edge'
+export const alt = 'Image of politician and their stance on crypto'
+export const size = {
+  width: 1200,
+  height: 630,
+}
+
+export const contentType = 'image/png'
+
+export default generateOgImage
```
Comment (olavoparno) [inline src/app/au/politicians/person/[dtsiSlug]/opengraph-image.tsx:13]: maybe add all of these exports in a file inside common as well and simply import as `import '@/components/app/pagePoliticianDetails/common/ogImageDetails'`, would something like that work?

src/app/[countryCode]/politicians/page.tsx
```diff
@@ -25,8 +25,8 @@ export async function generateMetadata(_props: Props): Promise<Metadata> {
 export default async function PoliticiansHomepage(props: PageProps) {
   const params = await props.params
   const { countryCode } = params
```
Comment (lucasrmp-cb) [inline src/app/[countryCode]/politicians/page.tsx:27]: Let's make this static on the page as a const (outside of the component)

`const countryCode = DEFAULT_SUPPORTED_COUNTRY_CODE`

src/app/[countryCode]/politicians/person/[dtsiSlug]/page.tsx
```diff
@@ -69,13 +42,13 @@ export default async function PoliticianDetails(props: Props) {
   const { countryCode } = params
```
Comment (lucasrmp-cb) [inline src/app/[countryCode]/politicians/person/[dtsiSlug]/page.tsx:42]: Same

src/app/au/politicians/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default async function PoliticiansHomepage() {
+  const results = await queryDTSIAllPeople({ countryCode: SupportedCountryCodes.AU })
+
+  const politicians = sortDTSIPersonDataTable(results.people).slice(0, 100)
+  return <AuPagePoliticians {...{ politicians }} />
```
lucasrmp-cb, line 30: np: why the spread, I think it makes it harder to read for no reason
```suggestion
  return <AuPagePoliticians politicians={politicians} />
```

src/app/au/politicians/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export async function generateMetadata(_props: Props): Promise<Metadata> {
+  return generateMetadataDetails({
+    title: PAGE_POLITICIANS_TITLE,
+    description: PAGE_POLITICIANS_DESCRIPTION,
+  })
+}
+
+export default async function PoliticiansHomepage() {
+  const results = await queryDTSIAllPeople({ countryCode: SupportedCountryCodes.AU })
```
lucasrmp-cb, line 27: np: let's keep the pattern of having the countryCode defined statically at the top of the file, my idea with that (besides having to change less places) is for the eng to know which country this relates at the first glance of the file

src/components/app/dtsiClientPersonDataTable/common/table.tsx
lucasrmp-cb, line 1: Can we separate components that are on this file a little bit? Don't need the "one per file" rule but at least by composition

src/components/app/dtsiClientPersonDataTable/common/table.tsx
```diff
@@ -0,0 +1,382 @@
...
+}
+
+export function DataTable({ children }: { children: ReactNode }) {
+  return <div className="space-y-6">{children}</div>
+}
+
+DataTable.GlobalFilter = DataTableGlobalFilter
+DataTable.Body = DataTableBody
+
+export function DataTableGlobalFilter({
```
lucasrmp-cb, line 68: Also, because you're also exporting these components I'd make sure that for clarity I'd put the root assignment (e.g.: `DataTable.GlobalFilter = DataTableGlobalFilter`) at the definition of the component and not at the root, so these lines would go below the component's definition

lucasrmp-cb: And on that note, why are we exporting these?

src/components/app/dtsiClientPersonDataTable/us/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+} from '@/utils/shared/usStateUtils'
+
+const GLOBAL_SEARCH_PLACEHOLDER = 'Search by name or state'
+const GLOBAL_SUBTITLE =
+  'We have a database of over 1,000 politicians. Search any politician to see where they stand on crypto.'
+const GLOBAL_TITLE = 'Search for a politician'
+
+export function UsDTSIClientPersonDataTable({
+  initialData,
+  countryCode,
```
lucasrmp-cb, line 44: Let's make this static, this component is US-specific

lucasrmp-cb (changes requested):

src/components/app/pageHome/au/index.tsx
```diff
@@ -97,51 +98,65 @@ export function AuPageHome({
...
+                  />
+                ))}
+              </DTSIPersonHeroCardRow>
+            </div>
+          )}
+          <div className="container space-x-4 text-center">
+            <Button asChild variant="secondary">
+              <InternalLink href={urls.politiciansHomepage()}>View all</InternalLink>
+            </Button>
+          </div>
```
lucasrmp-cb, line 157: Is there a way to extend this `HomeSection` component for use on this specifically? This is a bit much styling repetition IMO
Don't want a giant component that receives all the children, just something like a `PoliticiansHomeSection` that adds some more blocks with all the container and spacing styles and the different auxiliary div elements

src/components/app/pagePoliticianDetails/common/politiciansDetails.tsx
```diff
@@ -0,0 +1,174 @@
...
+          </ExternalLink>
+        </Button>
+      ))}
+    </div>
+  )
+}
+
+PagePoliticianDetails.Stances = PoliticianStances
+
+export function PoliticianStances({
```
lucasrmp-cb, line 144: Same here, why are we exporting composable components? They should only be used through `<PagePoliticianDetails.Stances />` (and if they are used outside the politicians details page the component name should be `<PoliticianDetails.Stances />` instead (see `recentActivity` as an example)

src/components/app/pagePoliticians/common/layout.tsx
```diff
@@ -0,0 +1,16 @@
...
+    <section className="standard-spacing-from-navbar container mb-16 space-y-7">{children}</section>
+  )
+}
+
+function PoliticiansTableSection({ children }: { children: React.ReactNode }) {
+  return <section>{children}</section>
+}
+
+PagePoliticiansLayout.IntroductionSection = IntroductionSection
+PagePoliticiansLayout.PoliticiansTableSection = PoliticiansTableSection
```
lucasrmp-cb, line 16: np: Let's keep consistency with where these assignments happen. Other places are already doing after the component's definition, let's do the same here

src/data/dtsi/queries/queryDTSIAllPeople.ts
```diff
@@ -32,14 +20,21 @@ const query = /* GraphQL */ `
 /*
 Because this request returns so many results, we should ensure we're only triggering this logic in cached endpoints/routes
 */
-export const queryDTSIAllPeople = async ({ limit }: { limit: number } = { limit: 1500 }) => {
+export const queryDTSIAllPeople = async ({
+  limit = 1500,
+  countryCode = SupportedCountryCodes.US,
```
lucasrmp-cb, line 25: it's a neat, but keeps consistency with other implementations we already have

```suggestion
  countryCode = DEFAULT_SUPPORTED_COUNTRY_CODE,
```

lucasrmp-cb (changes requested):

src/app/au/politicians/person/[dtsiSlug]/opengraph-image.tsx
Comment (twistershark) [reply src/app/au/politicians/person/[dtsiSlug]/opengraph-image.tsx:13]: Not sure. Any idea on how we can test this locally? 

Didn't find any reference in https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image

src/components/app/dtsiClientPersonDataTable/common/table.tsx
twistershark: They were in separate files initially and when I moved all of them to the same file, I forgot to remove the export

lucasrmp-cb (approved):

olavoparno (approved):

PR #2066, opened 2025-03-27
feat: adds IP location bypass logic and new international redirect logic

### LINKED ISSUE #2045: Create a IP location bypass

twistershark:
closes #2045 

## What changed? Why?

This PR does the following changes:
- This PR refactors how we handle user access location country code to be more descriptive and easier to understand.
- It also updates some options regarding country code in the internal page
- This PR also creates a new documentation about how we handle everything related to country code, geogate, geolocation, etc

**P.S.: As we need to default the USER_ACCESS_LOCATION to something on dev environment, we are using the page's country code as a fallback for the IP location. This means that we will never see the geogate banner on dev environment, if we don't override the access location or if we don't set the env to force a specific country code.**

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/internationalRedirectHandler.test.ts
```diff
@@ -33,6 +35,13 @@ const createMockRequest = (
     headers.set('cookie', `${USER_COUNTRY_CODE_COOKIE_NAME}=${options.countryCodeCookie}`)
   }
 
+  if (options?.userSelectedCountryCookie) {
+    headers.set(
+      'cookie',
+      `${USER_SELECTED_COUNTRY_COOKIE_NAME}=${options.userSelectedCountryCookie}`,
+    )
```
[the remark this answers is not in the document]
twistershark: This scenario is not relevant for this test suit.

olavoparno (approved):

src/utils/server/internationalRedirectHandler.ts
```diff
@@ -97,6 +113,23 @@ function shouldRedirectToCountrySpecificHomepage(
   return true
 }
 
+function shouldRedirectBasedOnUserSelectedCountry(
+  request: NextRequest,
+  userSelectedCountryCookie?: string,
+): boolean {
+  if (!userSelectedCountryCookie) return false
+
+  const isUShomepageRequested = new RegExp(`^/$`).test(request.nextUrl.pathname)
```
LucasdRossi, line 122: Is this Regex considering query params?

twistershark: I don't think so. Good call out

lucasrmp-cb (approved):

src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx
```diff
@@ -0,0 +1,95 @@
...
+    Cookies.remove(LOCAL_USER_CURRENT_SESSION_KEY)
+    Cookies.remove(USER_SESSION_ID_COOKIE_NAME)
+    Cookies.remove(USER_SELECTED_COUNTRY_COOKIE_NAME)
+
+    toast.success(
+      'IP location country code bypassed. Please, open a new tab to see the changes or refresh the page.',
+      {
+        duration: 10000,
+      },
+    )
```
Comment (olavoparno) [inline src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx:43]: maybe add a router.refresh() here?

src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const handleLocationCountryCodeReset = () => {
+    Cookies.remove(OVERRIDE_USER_ACCESS_LOCATION_COOKIE_NAME)
+    Cookies.remove(USER_ACCESS_LOCATION_COOKIE_NAME)
+
+    toast.success(
+      'IP location country code bypass removed! Please, open a new tab to see the changes or refresh the page.',
+      {
+        duration: 10000,
+      },
+    )
```
Comment (olavoparno) [inline src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx:55]: same for the router.refresh() here too

src/components/app/navbar/navbarCountrySelect.tsx
```diff
@@ -80,6 +88,12 @@ export function NavbarCountrySelect() {
               className="cursor-pointer data-[disabled]:font-bold data-[disabled]:text-primary-cta data-[disabled]:opacity-100"
               disabled={option.value === currentOption.value}
               key={option.value}
+              onClick={() => {
+                // This is necessary to bypass the redirect in the middleware, as if we don't remove the selected country cookie, the middleware will redirect the user to the current selected country, preventing users from switching to US.
+                if (option.value === DEFAULT_SUPPORTED_COUNTRY_CODE) {
```
olavoparno, line 93: that's because middleware runs before this component's useEffect, right?

olavoparno: we could maybe handle everything in a single onClick instead of having this onClick here and the link below

src/components/app/navbarGlobalBanner/currentCountryCampaignBannerContent.tsx
```diff
@@ -0,0 +1,100 @@
...
+function UKCampaign() {
+  return null
+}
+
+function CACampaign() {
+  return null
+}
+function AUCampaign() {
+  return null
+}
```
olavoparno, line 70: np: we prob want these on localized folders but since they're so little, we ought have them here anyways. just wondering

src/components/app/navbar/navbarCountrySelect.tsx
```diff
@@ -80,13 +86,15 @@ export function NavbarCountrySelect() {
               className="cursor-pointer data-[disabled]:font-bold data-[disabled]:text-primary-cta data-[disabled]:opacity-100"
               disabled={option.value === currentOption.value}
               key={option.value}
+              onClick={() => {
+                Cookies.set(USER_SELECTED_COUNTRY_COOKIE_NAME, option.value)
+                window.location.href = getIntlUrls(option.value).home()
```
twistershark, line 91: There was an unexpected bug with Next link and router trying to go back to us(/). Tried a bunch of different things and all failed. 

Decided to follow the pattern in `navbarGlobalBanner` and redirect using window.location.href to make sure that when switching countries, the experience will be reloaded to match the selected country

src/app/[countryCode]/internal/user-settings/updateUserAccountCountryCode.tsx
```diff
@@ -0,0 +1,71 @@
...
+  SupportedCountryCodes,
+} from '@/utils/shared/supportedCountries'
+
+export function UpdateUserAccountCountryCode() {
+  const router = useRouter()
+  const { isLoggedIn, user } = useSession()
+
+  const handleCountryCodeSubmit = async (countryCode: SupportedCountryCodes) => {
+    const result = await actionUpdateUserCountryCode(
+      countryCode.toLowerCase() as SupportedCountryCodes,
```
Comment (lucasrmp-cb) [inline src/app/[countryCode]/internal/user-settings/updateUserAccountCountryCode.tsx:24]: np: redundant type cast and `.toLowerCase` call

src/utils/server/internationalRedirectHandler.ts
```diff
@@ -1,119 +1,98 @@
...
+  maybeExistingUserAccessLocationCookie,
+  maybeUserSelectedCountryCookie,
+}: {
+  request: NextRequest
+  userAccessLocation: string
+  maybeExistingUserAccessLocationCookie?: string
+  maybeUserSelectedCountryCookie?: string
+}) {
+  // On local development, we want to bypass the international redirect if the BYPASS_INTERNATIONAL_REDIRECT environment variable is set to true
+  if (process.env.BYPASS_INTERNATIONAL_REDIRECT === 'true') {
```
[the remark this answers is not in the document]
lucasrmp-cb: > we should prob use toBool here to avoid issues

Not sure I agree, can you expand on the possible issues?

src/utils/server/middleware/setResponseCookie.ts
lucasrmp-cb, line 1: Weird comment but let me explain: I'd create a `src/utils/edge` to put this into. Middleware runs in a different env and the utils we can use for that env are different than the ones we use for server, having a different folder makes it clear that all functions should be [edge compliant](https://nextjs.org/docs/app/api-reference/edge#unsupported-apis) (it has a limited amount of apis and features)

src/utils/server/serverAnalytics/serverAnalytics.ts
```diff
@@ -92,15 +89,14 @@ export function getServerAnalytics(config: ServerAnalyticsConfig) {
       let countryCode: string | null = null
 
       try {
-        // We already have a function that gets the country code from the cookies (getCountryCodeCookie.ts)
+        // We already have a function that gets the country code from the cookies (getUserAccessLocationCookie.ts)
```
LucasdRossi, line 92: Can't you use a try/catch instead of duplicating the logic?

src/utils/shared/supportedCountries.ts
```diff
@@ -27,4 +27,11 @@ export const COUNTRY_CODE_REGEX_PATTERN = new RegExp(
   `^(${Object.values(SupportedCountryCodes).join('|')})$`,
 )
 
+export const COUNTRY_CODE_TO_DISPLAY_NAME: Record<SupportedCountryCodes, string> = {
+  [SupportedCountryCodes.US]: 'the United States',
+  [SupportedCountryCodes.GB]: 'the United Kingdom',
+  [SupportedCountryCodes.CA]: 'Canada',
+  [SupportedCountryCodes.AU]: 'Australia',
+}
```
lucasrmp-cb, line 35: Duplicate from `utils/shared/intl/displayNames`

lucasrmp-cb (changes requested):

src/utils/server/serverAnalytics/serverAnalytics.ts
twistershark: No, because the getUserAccessLocationCookie function validates the cookie value and if it is not supported, we throw an error. Using it would result in the function returning nothing because on edge env we won't have access to the cookie.

That's why we added a direct cookie access here. We dont care if it is valid, we just want to log it if it is available

src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx
Comment (twistershark) [reply src/app/[countryCode]/internal/user-settings/userAccessLocationBypass.tsx:55]: We need the request to hit the middleware again to generate analytics, session id and other necessary cookies. I'm not sure if router.refresh would be sufficient

src/utils/server/internationalRedirectHandler.ts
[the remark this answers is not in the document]
olavoparno: just because we can with the function and because we're using in similar cases throughout the project

olavoparno (approved):

LucasdRossi (approved):

PR #2068, opened 2025-03-27
feat: adds politicians section in homepage

### LINKED ISSUE #2064: Add politicians section to CA, AU and GB homepages

twistershark:
closes #2064 

## What changed? Why?

This PR adds new politicians section in CA, AU and GB homepages.
P.S.: As the only politician that is present in DTSI staging doesn't have any stances, we don't have any politicians to show in the homepage.
P.S.2: The politicians page is created in this PR: https://github.com/Stand-With-Crypto/swc-web/pull/2060

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/dtsiPersonHeroCard/index.tsx
```diff
@@ -167,7 +170,7 @@ export function DTSIPersonHeroCard(props: Props) {
               )}
             </div>
             <div className="ml-auto h-12 w-10 flex-shrink-0">
-              <DTSIFormattedLetterGrade className="h-full w-full" person={person} />
+              {cloneElement(cryptoStanceGradeComponent, { className: 'h-full w-full' })}
```
eduardopicolo-cb: It would be better to refactor the `DTSIPersonHeroCard` into composition, it would eliminate the need for `cloneElement`, which is generally considered an anti-pattern as it makes the code harder to understand and maintain

Example usage:
<DTSIPersonHeroCard>
  <DTSIPersonHeroCard.Image />
  <DTSIPersonHeroCard.Grade />
  <DTSIPersonHeroCard.Footer />
</DTSIPersonHeroCard>

olavoparno (approved):

src/components/app/dtsiThumbsUpOrDownGrade/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+      return '/dtsiLetterGrade/thumbs-up.svg'
+    }
+    if (isAntiCrypto(letterGrade)) {
+      return '/dtsiLetterGrade/thumbs-down.svg'
+    }
+    return '/dtsiLetterGrade/no-stance.svg'
+  }
+  return (
+    // eslint-disable-next-line @next/next/no-img-element
+    <img alt="Politician's crypto stance" className={className} src={getGradeImage()} />
```
eduardopicolo-cb, line 38: Why are we not using `NextImage` here?
Maybe a switch case for each grade would allow for a static `src`

eduardopicolo-cb (changes requested):

eduardopicolo-cb:
type check is breaking in some files

src/data/dtsi/queries/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+  [SupportedCountryCodes.AU]: [
+    DTSI_PersonGrouping.CURRENT_AU_HOUSE_OF_REPS,
+    DTSI_PersonGrouping.RUNNING_FOR_AU_HOUSE_OF_REPS,
+    DTSI_PersonGrouping.NEXT_AU_HOUSE_OF_REPS,
+  ],
+  [SupportedCountryCodes.GB]: [
+    DTSI_PersonGrouping.CURRENT_UK_HOUSE_OF_COMMONS,
+    DTSI_PersonGrouping.NEXT_UK_HOUSE_OF_COMMONS,
+    DTSI_PersonGrouping.RUNNING_FOR_UK_HOUSE_OF_COMMONS,
+  ],
```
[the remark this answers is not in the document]
twistershark: Updated! I think DTSI updated the schema again after my first implementation

src/components/app/dtsiThumbsUpOrDownGrade/index.tsx
twistershark: I just replicated what was implemented in the Letter grade component.

I think the reason why img was used instead of NextImage is because we pass width and height as tailwind classname, which allows us to have dynamic height and width for mobile and desktop.

If we change this component to accept height and width instead of classname, we would need to change across the entire SWC codebase to repeatedly pass desktop as w-full h-full and mobile as w-5 h-5

src/data/dtsi/queries/queryDTSIHomepagePeople.ts
```diff
@@ -1,19 +1,41 @@
...
   ${fragmentDTSIPersonCard}
 `
 
-export const queryDTSIHomepagePeople = () => {
-  return fetchDTSI<DTSI_HomepagePeopleQuery, DTSI_HomepagePeopleQueryVariables>(query)
+export const queryDTSIHomepagePeople = ({
+  countryCode = SupportedCountryCodes.US,
+}: {
+  countryCode?: string
+} = {}) => {
```
lucasrmp-cb, line 33: Let's type this correctly, it should be a `SupportedCountryCode`, we shouldn't be casting random strings to it

lucasrmp-cb (changes requested):

src/data/dtsi/queries/queryDTSIHomepagePeople.ts
twistershark: Updated!

lucasrmp-cb (approved):

eduardopicolo-cb (approved):

PR #2106, opened 2025-04-01
fix: intl bug bash items

lucasrmp-cb:
closes 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/getCountryCodeCookie.ts
```diff
@@ -35,7 +36,7 @@ export async function getCountryCodeCookie({
       throw error
     }
 
-    return countryCode
+    return countryCode as TBypassCountryCodeCheck extends true ? string : SupportedCountryCodes
```
LucasdRossi, line 39: np: move this to a type since you're using it two times

lucasrmp-cb: Actually makes it more convoluted, bc the only way to do it is add to the type args, the problem is that it needs to be defined after `TBypassCountryCodeCheck` (because it depends on it), it can't be defined after without a default value (bc a required arg can't be set after a non required one) and adding a default value to that breaks the `SupportedCountryCodes` cast on the last return. Fixing that actually would make this borderline unreadable

LucasdRossi (approved):

Issue #2102, opened 2025-04-01
Modify Key Races display flow for the US


olavoparno:
We'll want to [modify the key races flow](https://www.figma.com/design/8bNIcCkAoCO6nhcimN8teH/SWC-Templates?node-id=19-109) on the vote page based off this requirement

olavoparno:
- [ ] whoever is in office, don't hide their info in politicians/politicians details pages
- [ ] for everyone else, hide their scores on /races pages
- [ ] persist party selection across /races pages

PR #2107, opened 2025-04-01
Refactor/races UI

olavoparno:
closes  https://github.com/Stand-With-Crypto/swc-web/issues/2102

fixes 

## What changed? Why?

https://swc-web-git-refactor-races-ui-coinbase-vercel.vercel.app/races

image

## How has it been tested?

- [X] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/dtsiStanceDetails/index.tsx
```diff
@@ -10,20 +10,23 @@ import {
...
-import { isPoliticianStanceHidden } from '@/utils/dtsi/dtsiPersonUtils'
 import { convertDTSIStanceScoreToCryptoSupportLanguage } from '@/utils/dtsi/dtsiStanceScoreUtils'
 import { cn } from '@/utils/web/cn'
 
-function StanceTypeContent({ stance: passedStance, ...props }: IStanceDetailsProps) {
+function StanceTypeContent({
+  stance: passedStance,
+  isStanceHidden,
+  ...props
+}: IStanceDetailsProps) {
```
twistershark, line 20: Np: Your PR didn't create this, but we don't use interfaces with I prefix

src/components/ui/responsiveTabsOrSelect.tsx
```diff
@@ -11,26 +12,52 @@ import {
...
+    setCurrentTab(value)
+
+    if (persistCurrentTab) {
+      persistCurrentTabInCookies(value)
+    }
+  }
+
+  useEffect(() => {
+    if (persistCurrentTab) {
+      setCurrentTab(getCurrentTabFromCookies(defaultValue))
```
twistershark, line 52: Would it be possible to move this set to the useState initialization? You could change it from default value to a function that returns de persisted or the default value. Would that create any problems on the component?

twistershark (approved):

src/components/ui/responsiveTabsOrSelect.tsx
olavoparno: has to follow the lifecycle otherwise we get hydration problems

PR #2123, opened 2025-04-03
feat: add view_key_page and get informed actions

lucasrmp-cb:
closes 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/userActionGridCTAs/constants/gb/gbCtas.tsx
```diff
@@ -59,6 +62,28 @@ export const GB_USER_ACTION_CTAS_FOR_GRID_DISPLAY: UserActionGridCTA = {
       },
     ],
   },
+  [UserActionType.VIEW_KEY_PAGE]: {
```
LucasdRossi, line 65: What will happen if more then one `VIEW_KEY_PAGE` is active?

lucasrmp-cb: It will go to the modal for multiple campaigns, this implementation doesn't support that and it's not the idea (for simplicity)

LucasdRossi (approved):

PR #2196, opened 2025-04-14
disable maybeGetCongressionalDistrictFromAddress for intl domains

### LINKED ISSUE #2149: disable `maybeGetCongressionalDistrictFromAddress` on `actionUpdateUserProfile` for intl websites
Currently, all addresses filled in update profile dialog will trigger a US congressional district lookup on that address, which will always fail for users outside of US.

As we as starting to support more countries, we don't want this false positive error to flood our monitoring systems.

eduardopicolo-cb:
closes #2149 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/actions/actionUpdateUserProfile.ts
+++ b/src/actions/actionUpdateUserProfile.ts
@@ -67,8 +67,9 @@ async function actionUpdateUserProfileWithoutMiddleware(
     }
   }
 
+  const isUSAddress = validatedFields.data.address?.countryCode?.toUpperCase() === 'US'
   try {
-    if (validatedFields.data.address) {
+    if (isUSAddress && validatedFields.data.address) {
       const usCongressionalDistrict = await maybeGetCongressionalDistrictFromAddress(
         validatedFields.data.address,
       )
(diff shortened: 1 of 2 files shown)


src/actions/actionUpdateUserProfile.ts
```diff
@@ -67,8 +67,9 @@ async function actionUpdateUserProfileWithoutMiddleware(
     }
   }
 
+  const isUSAddress = validatedFields.data.address?.countryCode?.toUpperCase() === 'US'
```
twistershark, line 70: np: I would use `DEFAULT_SUPPORTED_COUNTRY_CODE.toUpperCase()` instead of `'US'` just to keep the code consistent

src/utils/server/externalOptIn/handleExternalUserActionOptIn.ts
```diff
@@ -322,27 +322,30 @@ async function maybeUpsertUser({
       logger.error('error getting `googlePlaceId`:' + e)
     }
 
-    try {
-      const usCongressionalDistrict = await maybeGetCongressionalDistrictFromAddress(dbAddress)
+    const isUSAddress = dbAddress.countryCode?.toUpperCase() === 'US'
```
twistershark, line 325: np: I would use `DEFAULT_SUPPORTED_COUNTRY_CODE.toUpperCase()` instead of `'US'` just to keep the code consistent

twistershark (approved):

PR #2221, opened 2025-04-23
Postgis config

lucasrmp-cb:
closes 

fixes 

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/data/postgis/getCongressionalDistrictData.ts
lucasrmp-cb, line 1: Let's move the full `swc-civic` library into `src/utils/shared`

.env.example
```diff
@@ -51,7 +51,7 @@ DATABASE_URL="mysql://root:root@localhost:3306/swc-web"
 
 # This is the URL for the PostGIS database that contains the district geo data
 # To start the PostGIS container, run `make dc_up`
-DISTRICT_GEO_DATABASE_URL="postgres://postgres:postgres@localhost:5432/swc_district_geo_data"
+DISTRICT_GEO_DATABASE_URL="postgres://postgres:postgres@localhost:5432/swc_civic"
```
lucasrmp-cb, line 54: If the db name changed we should change the env var name too (and the comment here, basically everywhere we mention this service as "district geo")

Makefile
lucasrmp-cb, line 1: Not sure I like having this file here, a `Makefile` in the root that controls a single small system doesn't make a lot of sense to me. My suggestion would be to move to `/bin` and change the file name to `swc_civic.mk` or something like that.

[This](https://web.mit.edu/gnu/doc/html/make_3.html#:~:text=If%20you%20want%20to%20use,you%20can%20specify%20several%20makefiles.) is how to do it in make

prisma-swc-civic/schema.prisma
```diff
@@ -0,0 +1,90 @@
...
+  begin        DateTime?                @db.Timestamptz(6)
+  end          DateTime?                @db.Timestamptz(6)
+  altitudemode String?                  @db.VarChar
+  tessellate   Int?
+  extrude      Int?
+  visibility   Int?
+  draworder    Int?
+  icon         String?                  @db.VarChar
+  snippet      String?                  @db.VarChar
+  wkb_geometry Unsupported("geometry")?
```
lucasrmp-cb, line 43: Can we specify encoding here? We are doing it in the raw queries (same to the other tables)
```suggestion
  wkb_geometry Unsupported("geometry(Geometry,4326)")?
```

lucasrmp-cb (commented):
Can't approve, but LGTM

PR #2227, opened 2025-04-25
SWC US Politician Updates

travisbloom-cb:
This PR:
- removes old President/VPs from showing up on politicians table
- Adds AZ-07 Congressional race to key races page
- Adds Governors/AGs to US poltiicians pages
- Fixes filters for politicians page to be strongly typed so we have type guarantees when we expand the filters we support
- Refactored the way we determine whether to show someones scores
   - for components on race pages, we explicitly define whether the component should show scores (because it depends on if the race is for a federal or state role)
   - for other components used generically elsewhere on the site, we leverage the person's primary role instead of hardcoding a bunch of slugs. This ensures we don't have bugs in the future as additional politicians are added to the site

## Follow up actions
In a separate PR, we should refactor the way we do filters in other locales to give us the same level of type safety

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


travisbloom-cb:
This is in draft until we confirm some items from SWC

olavoparno (approved):

src/utils/shared/locationSpecificPages.ts
twistershark, line 1: No idea what this file does and which problems this change could cause. @olavoparno, please verify if there is any problems here

twistershark (dismissed):

src/utils/shared/locationSpecificPages.ts
olavoparno: it's just being used to render districts people inside its state page

PR #2244, opened 2025-04-30
Sync Sendgrid contact list cron job

eduardopicolo-cb:
closes https://github.com/Stand-With-Crypto/swc-internal/issues/369

fixes 

## What changed? Why?

- Added Inngest cron job to sync intl swc users to Sendgrid's Contact Lists

## Notes to reviewers

Current implementation does not includes US users and does not supports large amounts of users, further optimization is needed. It is meant to support last minute demand to allow the intl team to send marketing emails.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/sendgrid/marketing/contacts.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+  })
+
+  const data = {
+    list_ids: options.listIds,
+    contacts: formattedContacts,
+  }
+
+  try {
+    const request = {
```
lucasrmp-cb, line 65: Let's type this

src/utils/server/sendgrid/marketing/contacts.ts
```diff
@@ -0,0 +1,218 @@
...
+      }
+    })
+
+    return {
+      ...standardFields,
+      custom_fields: customFieldsMappedToIds,
+    }
+  })
+
+  const data = {
```
lucasrmp-cb, line 59: Let's type this if possible

src/utils/server/sendgrid/marketing/contacts.ts
```diff
@@ -0,0 +1,218 @@
...
+          },
+        })
+      }
+    })
+
+    return {
+      ...standardFields,
+      custom_fields: customFieldsMappedToIds,
+    }
+  })
```
lucasrmp-cb, line 57: Can this be turned into a function?

src/utils/server/sendgrid/marketing/contacts.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+interface SendgridImportResponse {
+  job_id: string
+  upload_uri: string
+  upload_headers: Array<{ header: string; value: string }>
+}
+
+export async function uploadSendgridContactsCSV(contacts: SendgridContact[], options?: Options) {
```
lucasrmp-cb, line 97: Let's send this into it's own file and try to separate concerns, this function seems to be doing way too much stuff

src/utils/server/sendgrid/marketing/lists.ts
```diff
@@ -0,0 +1,63 @@
...
+    name: string
+    contact_count: number
+  }[]
+}
+
+interface CreateListResponse {
+  id: string
+  name: string
+  contact_count: number
+}
```
lucasrmp-cb, line 29: This is a `List` type isn't it? Two question based on it:
1. Doesn't the Sendgrid sdk provide these typings we're doing all around this PR?
2. We prob should name this better and use the same type on the `GetListsResponse` above

src/utils/shared/intl/displayNames.ts
```diff
@@ -1,11 +1,11 @@
 import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
 
-export const COUNTRY_CODE_TO_DISPLAY_NAME: Record<SupportedCountryCodes, string> = {
+export const COUNTRY_CODE_TO_DISPLAY_NAME = {
```
lucasrmp-cb, line 3: Why? This is removing the strict typing, if we add a new `SupportedCountryCodes` and don't update this it wont break this as it should

src/utils/server/sendgrid/marketing/lists.ts
eduardopicolo-cb: [Sendgrid sdk](https://github.com/sendgrid/sendgrid-nodejs/tree/main/packages/contact-importer) for Marketing is using the legacy API (`/contactdb` endpoints). Currently the only way to use the new Marketing API is to request the `/v3/marketing` endpoints.

The sendgrid lib currently does not provides any payload/response typings, other than a generic object with
```body: any; code: number;```

Im going to improve the types and namings.

File: src/utils/shared/intl/displayNames.ts (hunk 6 above)
Comment (eduardopicolo-cb) [reply src/utils/shared/intl/displayNames.ts]: Fixed

Needed so that the type `SendgridContactListName` resolves to:
```
"United States Advocates" | "Canada Advocates" | "United Kingdom Advocates" | "Australia Advocates"
```
instead of: 
```
"'string' Advocates"
```

lucasrmp-cb (approved):

PR #2268, opened 2025-05-05
chore: added au tracking scripts

luiz-from-delta:
closes #364

## What changed? Why?

Added Twitter and LinkedIn scripts for tracking in AU page layout.

## Notes to reviewers

The scripts should show up only for AU layout.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

**IMPORTANT:**
- Two new environment variables have been added: NEXT_PUBLIC_AU_TWITTER_EVENT_ID and NEXT_PUBLIC_AU_LINKEDIN_PARTNER_ID.


src/components/app/layoutCustomScripts.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const isEnvAllowed = Array.isArray(envs)
+    ? envs.includes(NEXT_PUBLIC_ENVIRONMENT)
+    : envs === NEXT_PUBLIC_ENVIRONMENT
+
+  return isEnvAllowed
+}
+
+export function getLayoutCustomScript(countryCode: SupportedCountryCodes) {
```
lucasrmp-cb, line 76: We don't do use jsx generator functions as a pattern in the project. Let's refactor this is a component with normal life cycle.

A way to organize it would be something like this:

```
- src/components/app/
  - layoutScripts/
    - au/index.tsx (this would contain the au specific script)
    - index.tsx (this would contain a component with the logic to render `au/index.tsx` or any other component conditionally)
```

lucasrmp-cb (changes requested):

lucasrmp-cb (approved):

twistershark (approved):

PR #2305, opened 2025-05-19
feat: created politicians section

### LINKED ISSUE #2265: Politicians section
Just the politicians query filtered by the state

luiz-from-delta:
closes #2265 

## What changed? Why?

Created a new section called Politicians to show the politicians filtered by state.

## How has it been tested?

- [ ] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


lucasrmp-cb (changes requested):

src/utils/shared/urls/index.ts
```diff
@@ -144,7 +144,7 @@ export const getIntlUrls = (
     politiciansHomepage: (filters?: Partial<{ state: string }>) => {
       const params = new URLSearchParams(filters).toString()
 
-      return `${countryPrefix}/politicians${params ? `?${params}` : ''}`
+      return `${countryPrefix}/politicians${params ? `?${params}#table` : ''}`
```
lucasrmp-cb, line 147: This doesn't look right, you can add a `hash` arg and just use it

```suggestion
    politiciansHomepage: ({ filters, hash }: { filters?: Partial<{ state: string }>, hash?: string } = {}) => {
      const params = new URLSearchParams(filters).toString()

      return `${countryPrefix}/politicians${params ? `?${params}` : ''}${hash ? `#${hash}` : ''}`
```

src/app/[countryCode]/local-policy/[stateCode]/page.tsx
```diff
@@ -20,10 +25,17 @@ export const metadata: Metadata = {
   }),
 }
 
+export function generateStaticParams() {
```
Comment (lucasrmp-cb) [inline src/app/[countryCode]/local-policy/[stateCode]/page.tsx:28]: We should also add `dynamic = "error"` on this page, and since all params are known and pre generated we should also set `dynamicParams = false`

src/components/app/pageLocalPolicy/common/statePage/politiciansSection.tsx
```diff
@@ -0,0 +1,29 @@
...
+  lowestScores,
+  stateCode,
+}: PoliticiansSectionProps) {
+  return (
+    <PoliticiansGrid
+      CryptoStanceGrade={DTSIFormattedLetterGrade}
+      countryCode={countryCode}
+      highestScores={highestScores}
+      lowestScores={lowestScores}
+      showGroupTitle={false}
```
twistershark, line 25: Nit picking: I'm not sure what this prop is used for, but does it make sense to have it hardcoded in the common folder? 

IMO maybe we add showGroupTitle as an optional prop and set the default value to false. But, I could be unnecessarily suggesting an abstracting that won't be used in the future, so ignore this if it doesn't make sense.

src/app/api/public/state-advocates/[stateCode]/route.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { getAdvocatesCountByStateData } from '@/data/pageSpecific/getAdvocatesCountByStateData'
+
+const zodPayload = z.object({
+  stateCode: z.string().length(2),
+})
+
+interface RequestContext {
+  params: Promise<{
+    stateCode: string
```
Comment (twistershark) [inline src/app/api/public/state-advocates/[stateCode]/route.ts:14]: Nit picking: I'm a bit worried that stateCode could be any string. It kind of suggests to the reader that it will work with any stateCode, not only US ones. Maybe we could make this more strict? Just a suggestion

PR #2354, opened 2025-06-09
Allow users re-subscribe to Sendgrid marketing emails

### LINKED ISSUE #2321: Allow users re-subscribe to Sendgrid marketing emails

eduardopicolo-cb:
closes #2321 

fixes 

## What changed? Why?

image

image

image

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/pageUserProfile/common/smsSubscriptionForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [isSubmitting, setIsSubmitting] = useState(false)
+
+  const hasOptedInToSMS = userHasOptedInToSMS(user)
+
+  const handleSMSOptInChange = async (smsOptIn: boolean) => {
+    setIsSubmitting(true)
+
+    const payload = {
+      phoneNumber,
+      optedInToSms: smsOptIn,
```
twistershark, line 34: np: just rename `smsOptIn` to `optedInToSms`. Will maintain the same meaning

src/components/app/pageUnsubscribe/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+        height={80}
+        priority
+        src="/logo/shield.svg"
+        width={80}
+      />
+
+      <section className="space-y-8">
+        <PageTitle size="md">Unsubscribed Successfully</PageTitle>
+
+        <PageSubTitle className="max-w-[600px]">
```
twistershark, line 24: np: it's better to use Tailwind tokens instead of hardcoded pixel numbers to make it more consistent and easier to maintain. I noticed on the [docs](https://tailwindcss.com/docs/width) that there is we have `w-xl` that represents `576px` and `w-2xl` that represents `672px`. Can you try both and check if any of these works?

src/components/app/pageUnsubscribe/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+          Please <span className="text-primary-cta">resubscribe</span> if you've changed your mind.
+        </PageSubTitle>
+
+        <SuspenseResubscribeButton />
+      </section>
+
+      <section className="space-y-4 text-center">
+        <SocialIcons />
+        <div>
+          <InternalLink href={getIntlUrls(SupportedCountryCodes.US).home()}>
```
twistershark, line 36: Same as the other comment here.

src/components/app/pageUnsubscribe/resubscribeButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+        payload: emailAddress,
+      },
+      removeFromGlobalSuppressionGroup,
+    )
+
+    if (result.status === 'success') {
+      toast.success('Successfully resubscribed to our mailing list!')
+      router.push('/embedded/email/resubscribe-success')
+    } else {
+      toastGenericError()
```
twistershark, line 46: Can't you just use the `onError` prop on `triggerServerActionForForm`?

```jsx
onError: toastGenericError,
```

src/components/app/pageUserProfile/common/emailSubscriptionForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+        emailOptIn
+          ? 'Successfully subscribed to our emails!'
+          : 'Successfully unsubscribed from our emails!',
+      )
+    }
+
+    setIsSubmitting(false)
+  }
+
+  const helpText = useMemo(() => {
```
twistershark, line 68: Do you really need a useMemo for this?

IMO, just using `const helpText = !emailAddress ? 'Please provide an email address to subscribe to our emails.' : ''`, would be enough

src/components/app/pageUserProfile/common/emailSubscriptionForm.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface EmailSubscriptionFormProps {
+  user: PageUserProfileUser
+}
+
+export function EmailSubscriptionForm({ user }: EmailSubscriptionFormProps) {
+  const { primaryUserEmailAddress } = user
+  const emailAddress = primaryUserEmailAddress?.emailAddress ?? ''
+
+  const [isSubmitting, setIsSubmitting] = useState(false)
+  const isFirstLoad = useRef(true)
```
twistershark, line 28: Instead of using this isFirstLoad logic in this component, have you tried to use suspense?

You could wrap this component in it's parent like this:
```tsx
 <Suspense fallback={<Skeleton className="h-4 w-4" />}>
    <EmailSubscriptionForm user={user} />
  </Suspense>
```

Then, you would be able to remove the isFirstLoad related code.

I'm suggesting this as I imagine that this is being used to workaround a hydration error or something like that. I didn't test this suggestion I'm doing, but it might work and would make the code simpler

src/components/app/pageUnsubscribe/socialIcons.tsx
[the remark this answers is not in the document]
eduardopicolo-cb: Its for all geos. But since the user is coming from the link setup in Sendgrid, we don't have any way get the country code (without significant changes to the middleware). I brought this to our PM's attention, and most likely we are going to remove the social links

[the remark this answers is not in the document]
twistershark: On each page visit, our middleware already handles creating access location cookie if necessary. Can't we use it to check which country we should redirect the user? I'm saying this because I remember seeing a link to the US homepage on the SWC logo

twistershark (approved):

PR #2367, opened 2025-06-12
feat: created view key page redirect action

### LINKED ISSUE #2358: Create vew_key_page redirect action page
What we want to build here is an action deeplink for the `VIEW_KEY_PAGE` action (the url should be `/action/view-key-page`, this deeplink should receive the following params:

- `campaignName`: this should map exactly to the campaign name, if no match is found on the frontend than we never go to the server action

What the deeplink will do is:
1. Check for a valid `campaignName` and error out if it's invalid (with a `notFound()`)
2. Show a full screen loading spinner (see the email action modal for reference)
3. Trigger a server action `actionCreateUserActionViewKeyPage`
4. When the server action resolves we should redirect the user to the arbitrary url we need as an info when creating the campaign
    - This should have a minimum wait time of 1 second to avoid flashing a loading screen and worsening UX

Example:

We have a `GENIUS_VIDEO_TAKEOVER` campaign that is configured to point to `foobar.org` on the deeplink

Whe I access `standwithcrypto.org/action/view-key-page?campaignName=GENIUS_VIDEO_TAKEOVER` I want to see a loading screen (for at least one second) and then be redirected to `foobar.org`

luiz-from-delta:
closes #2358 

## What changed? Why?

Created a new redirect action for view key page.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/homepageDialogDeeplinkLayout/gb/index.tsx
+++ b/src/components/app/homepageDialogDeeplinkLayout/gb/index.tsx
@@ -14,11 +14,14 @@ import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
 
 const countryCode = SupportedCountryCodes.GB
 
+type GBHomepageDialogDeeplinkLayoutProps =
+  | ({ hidePseudoDialog?: false } & HomepageDialogDeeplinkLayoutProps)
+  | { hidePseudoDialog: true; children: React.ReactNode }
+
 export async function GBHomepageDialogDeeplinkLayout({
   children,
-  size = 'md',
-  className,
-}: HomepageDialogDeeplinkLayoutProps) {
+  ...props
+}: GBHomepageDialogDeeplinkLayoutProps) {
   const [topLevelMetrics, recentActivity, partners, founders, dtsiHomepagePoliticians] =
     await Promise.all([
       getHomepageTopLevelMetrics(),
@@ -33,9 +36,17 @@ export async function GBHomepageDialogDeeplinkLayout({
 
   return (
     <>
-      <PseudoDialog className={className} countryCode={countryCode} size={size}>
-        {children}
-      </PseudoDialog>
+      {props.hidePseudoDialog ? (
+        children
+      ) : (
+        <PseudoDialog
+          className={props.className}
+          countryCode={countryCode}
+          size={props.size || 'md'}
+        >
+          {children}
+        </PseudoDialog>
+      )}
 
       <GbPageHome
         dtsiHomepagePoliticians={dtsiHomepagePoliticians}
--- a/src/components/app/userActionViewKeyPageDeeplinkWrapper/gb.tsx
+++ b/src/components/app/userActionViewKeyPageDeeplinkWrapper/gb.tsx
@@ -1,0 +1,31 @@
+import { SearchParams } from 'next/dist/server/request/search-params'
+
+import {
+  CampaignMetadataMap,
+  UserActionViewKeyPageDeeplinkWrapper,
+} from '@/components/app/userActionViewKeyPageDeeplinkWrapper'
+import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
+import { GBUserActionViewKeyPageCampaignName } from '@/utils/shared/userActionCampaigns/gb/gbUserActionCampaigns'
+
+const countryCode = SupportedCountryCodes.GB
+
+const campaignMetadataMap: CampaignMetadataMap<keyof typeof GBUserActionViewKeyPageCampaignName> = {
+  NEWMODE_EMAIL_ACTION: {
+    url: 'https://www.google.com/',
+  },
+}
+
+export function GbUserActionViewKeyPageDeeplinkWrapper({
+  searchParams,
+}: {
+  searchParams: SearchParams | undefined
+}) {
+  return (
+    <UserActionViewKeyPageDeeplinkWrapper
+      campaignMetadataMap={campaignMetadataMap}
+      countryCode={countryCode}
+      pathName="/gb/action/view-key-page"
+      searchParams={searchParams}
+    />
+  )
+}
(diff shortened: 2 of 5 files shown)


twistershark (approved):

twistershark (dismissed):

src/components/app/homepageDialogDeeplinkLayout/gb/index.tsx
```diff
@@ -14,11 +14,23 @@ import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
...
+}
+
+type GBHomepageDialogDeeplinkLayoutProps =
+  | GBHomepageDialogDeeplinkLayoutWithDialog
+  | GBHomepageDialogDeeplinkLayoutWithoutDialog
+
 export async function GBHomepageDialogDeeplinkLayout({
   children,
-  size = 'md',
-  className,
```
lucasrmp-cb, line 20: Why replace names props with a spread? Stuff like the syntax for `size` is now more error prone than before

src/components/app/userActionViewKeyPageDeeplinkWrapper/loading.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  useEffect(() => {
+    const minWaitTimeTimeout = setTimeout(() => {
+      setStatus('ready')
+    }, minWaitTimeInSeconds * 1_000)
+
+    return () => {
+      clearInterval(minWaitTimeTimeout)
+    }
+  }, [minWaitTimeInSeconds])
```
lucasrmp-cb, line 47: np: I'd suggest separating this part into a custom hook, an api like this below would look way better to me
```suggestion
  useTimeout(() => setStatus('ready'), minWaitTimeInSeconds * 1000)
  // ...
  
  // bottom of the file
  function useTimeout(callback, ms) {
    useEffect(() => {
      const timeoutRef = setTimeout(callback, ms)

      return () => {
        clearTimeout(timeoutRef)
      }
    }, [callback, ms])
  }
```

src/components/app/userActionViewKeyPageDeeplinkWrapper/loading.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const [actionHasBeenCreated, setActionHasBeenCreated] = useState(false)
+
+  useEffect(() => {
+    const minWaitTimeTimeout = setTimeout(() => {
+      setStatus('ready')
+    }, minWaitTimeInSeconds * 1_000)
+
+    return () => {
+      clearInterval(minWaitTimeTimeout)
```
lucasrmp-cb, line 45: You're calling `clearInterval` in a timeout ref, this should be `clearTimeout`

src/components/app/userActionViewKeyPageDeeplinkWrapper/loading.tsx
```diff
@@ -0,0 +1,74 @@
...
+  campaignMetadata: CampaignMetadata
+  campaignName: string
+  countryCode: SupportedCountryCodes
+  minWaitTimeInSeconds?: number
+  pathName: string
+}
+
+const MIN_WAIT_TIME_IN_SECONDS = 1
+
+export function UserActionViewKeyPageDeeplinkLoading({
```
lucasrmp-cb, line 26: This name makes me think this is a UI only component, let's either change the name or separate some concerns here (fyi I'd describe this as a `Redirect` component)

src/components/app/userActionViewKeyPageDeeplinkWrapper/loading.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }, [minWaitTimeInSeconds])
+
+  useEffect(() => {
+    async function runServerAction() {
+      if (!actionHasBeenCreated) {
+        await actionCreateUserActionViewKeyPage({
+          campaignName,
+          countryCode,
+          path: pathName,
```
lucasrmp-cb, line 55: The path is what's saved as the page the user saw, it should be the `campaignMetadata.url`, not the deeplink url

lucasrmp-cb (changes requested):

src/components/app/userActionViewKeyPageDeeplinkWrapper/gb.tsx
```diff
@@ -0,0 +1,30 @@
...
+  CampaignMetadataMap,
+  UserActionViewKeyPageDeeplinkWrapper,
+} from '@/components/app/userActionViewKeyPageDeeplinkWrapper'
+import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
+import { GBUserActionViewKeyPageCampaignName } from '@/utils/shared/userActionCampaigns/gb/gbUserActionCampaigns'
+
+const countryCode = SupportedCountryCodes.GB
+
+const campaignMetadataMap: CampaignMetadataMap<keyof typeof GBUserActionViewKeyPageCampaignName> = {
+  NEWMODE_EMAIL_ACTION: {
```
lucasrmp-cb, line 13: This should use the campaign enum, not a hardcoded value
```suggestion
  [GBUserActionViewKeyPageCampaignName.NEWMODE_EMAIL_ACTION]: {
```

In a lot of cases the key might not equal the value exactly, would not be consistent with other maps throughout the app

src/components/app/userActionViewKeyPageDeeplinkWrapper/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import {
+  CampaignMetadata,
+  UserActionViewKeyPageDeeplinkRedirect,
+} from '@/components/app/userActionViewKeyPageDeeplinkWrapper/redirect'
+import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
+
+export type dynamic = 'error'
+
+export type CampaignMetadataMap<K extends Uppercase<string>> = Partial<Record<K, CampaignMetadata>>
```
lucasrmp-cb, line 12: Question, what is `Partial` doing here? I suggested partial on our conversation for cases where the key would be a campaign name enum, here you're using a string so it looks like `Partial` isn't doing anything

lucasrmp-cb (changes requested):

twistershark (approved):

PR #2429, opened 2025-06-26
fix: resolve React 19 readonly property errors in localStorage polyfill

LucasdRossi:
closes 

fixes [PROD-SWC-WEB-66Z](https://stand-with-crypto.sentry.io/issues/6565136440/events/7273f093f1354bf2be48bd5b0467438e/)

## What changed? Why?

This PR fixes the "Cannot assign to read only property" TypeErrors that started appearing in Sentry 2 months ago on iOS Safari, particularly in private browsing mode.

### Root Cause

- The errors began when React 19 was released (December 2024), which introduced stricter property handling
- Our localStorage polyfill was using `writable: false` in `Object.defineProperty`, making the polyfill read-only
- React 19's stricter behavior now throws errors when trying to assign to these read-only properties
- This only affects iOS Safari users in private browsing mode where localStorage is unavailable

### Changes Made

1. **localStorage polyfill**: Changed `writable: false` to `writable: true` for React 19 compatibility
2. **Enhanced security**: Added `Object.seal()` to prevent property modification while maintaining assignability
3. **Sentry error grouping**: Added readonly property errors to common error patterns for better monitoring

### Security Considerations

The `writable: true` change is safe because:

- Strong CSP policies prevent malicious script injection
- `Object.seal()` prevents modification of the polyfill object's properties
- The polyfill only affects browsers without localStorage support
- Defense-in-depth security model remains intact

## Notes to reviewers

The main changes are in:

- `src/components/app/overrideGlobalLocalStorage.tsx` - Fix the localStorage polyfill
- `sentry.client.config.ts` - Add error grouping for better monitoring

This addresses a few errors occurring primarily on iOS Safari devices. The fix maintains security while ensuring React 19 compatibility.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/overrideGlobalLocalStorage.tsx
```diff
@@ -38,13 +38,18 @@ export function OverrideGlobalLocalStorage() {
...
+          window.localStorage = localStorageNoop
+        } catch (directAssignError) {
+          // If direct assignment fails, use defineProperty but with writable: true
+          Object.defineProperty(window, 'localStorage', {
+            value: localStorageNoop,
+            configurable: true,
+            enumerable: true,
+            writable: true,
+          })
+        }
```
lucasrmp-cb, line 52: Wish I didn't have a np for the first PR fully made with AI

Let's create an aux function to do this assign, exact same code block from line 41
```suggestion
        setWindowLocalStorage(localStorageNoop)
        // ...
        setWindowLocalStorage(sessionStorage)
```

lucasrmp-cb (approved):

PR #2435, opened 2025-06-27
fix: error state load on locale mismatch

### LINKED ISSUE #2335: Error state load on locale mismatch.
I have a Canadian (non-US address) and its populating district data on the US pages when I access it.

Image

---

### Expected behavior:

* For invalid addresses we should display `N/A` on the `District ranking` field
* For non-US users with invalid addresses we should change the message to something that notes the country dynamic (@matthewlow-cb for copy)

nrenard:
closes #2335 

## What changed? Why?

- Fixed an issue where non-US (e.g., Canadian) addresses were incorrectly showing US district data on US pages.
- Now, for invalid or non-US addresses, the district ranking displays "N/A" and the message is updated to clarify the address is outside the US.

## How has it been tested?

- [X] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/pageReferrals/yourDistrictRank.tsx
+++ b/src/components/app/pageReferrals/yourDistrictRank.tsx
@@ -73,7 +73,7 @@ function YourDistrictRankContent(props: YourDistrictRankContentProps) {
           value={address === 'loading' ? null : address}
         />
         <p className="pl-4 text-sm text-fontcolor-muted">
-          District rank not found, please try a different address.
+          Looks like your address is outside the U.S., so it's not part of any district here.
         </p>
       </div>
     )
--- a/src/components/app/pageReferrals/referralsCounter.tsx
+++ b/src/components/app/pageReferrals/referralsCounter.tsx
@@ -92,7 +92,7 @@ export function UserDistrictRank({ className }: { className?: string }) {
     }
 
     if (!rank || !districtRankingResponse.data) {
-      return null
+      return <p>N/A</p>
     }
 
     return (


src/components/app/pageReferrals/yourDistrictRank.tsx
```diff
@@ -46,15 +47,42 @@ interface YourDistrictRankContentProps {
...
   const districtRankingResponse = useGetDistrictRank({
     stateCode,
     districtNumber,
   })
 
-  if (districtRankingResponse.isLoading) {
+  const { data } = districtRankingResponse
+
+  const checkIfUSAddress = useCallback(async () => {
+    // Check if the address is still loading, if the address is not yet loaded, or if stateCode is already present.
```
twistershark, line 63: np: I don't think this comment is necessary

twistershark (dismissed):

src/components/app/pageReferrals/yourDistrictRank.tsx
```diff
@@ -46,15 +47,40 @@ interface YourDistrictRankContentProps {
...
+    setIsUSAddress('loading')
+
+    const addressDetails = await convertGooglePlaceAutoPredictionToAddressSchema(address)
+
+    if (addressDetails.countryCode.toLowerCase() === SupportedCountryCodes.US) {
+      setIsUSAddress(true)
+      return
+    }
+
+    setIsUSAddress(false)
```
lucasrmp-cb, line 74: np
```suggestion
    setIsUSAddress(addressDetails.countryCode.toLowerCase() === SupportedCountryCodes.US)
```

src/components/app/pageReferrals/yourDistrictRank.tsx
```diff
@@ -46,15 +47,40 @@ interface YourDistrictRankContentProps {
...
 
-  if (districtRankingResponse.isLoading) {
+  const { data } = districtRankingResponse
+
+  const checkIfUSAddress = useCallback(async () => {
+    if (isLoadingAddress || !address || stateCode) return
+
+    setIsUSAddress('loading')
+
+    const addressDetails = await convertGooglePlaceAutoPredictionToAddressSchema(address)
```
lucasrmp-cb, line 67: Since this is not critical for the UJ (just changing an error message) let's suppress this error and just set the `isUSAddress` variable as `true`

src/components/app/pageReferrals/yourDistrictRank.tsx
```diff
@@ -63,18 +89,18 @@ function YourDistrictRankContent(props: YourDistrictRankContentProps) {
...
   }
 
-  const { data } = districtRankingResponse
   if (!data) {
+    let message = 'District rank not found, please try a different address.'
+
+    if (!isUSAddress) {
+      message =
+        "Looks like your address is outside the U.S., so it's not part of any district here."
+    }
```
lucasrmp-cb, line 98: np: this could be a ternary (maybe even inline but I'm ok with separating from render)
```suggestion
    const invalidAddressErrorMessage = isUSAddress
      ? 'District rank not found, please try a different address.'
      :  "Looks like your address is outside the U.S., so it's not part of any district here."
```

PR #2577, opened 2025-08-07
fix: bill timeline bug bash

luiz-from-delta:
This pull request addresses the issues identified during the recent bug bash session. The following changes have been implemented to enhance the UI and improve functionality:

Alignment Adjustment:
- The “X” marker has been moved up by 1px to ensure better alignment with the through line.

Color Update:
- The color of the “X” marker has been changed from red to gray. This adjustment aims to reduce visual distraction and create a more cohesive design.

Hover and Click Behavior:
- Resolved the weird behavior experienced when clicking and hovering over small points within the timeline section. The interaction should now be smoother and more intuitive.

Outline Color Matching:
- The outline color of the empty plot points has been updated to match the color of the through line, providing a more uniform appearance.

### DIFF
--- a/src/components/ui/timeline/majorMilestone.tsx
+++ b/src/components/ui/timeline/majorMilestone.tsx
@@ -23,6 +23,12 @@ interface MajorMilestoneProps {
   milestone: Milestone
 }
 
+enum MajorMilestoneStatus {
+  NOT_ENABLED_OR_FAILED = 'NOT_ENABLED_OR_FAILED',
+  NOT_HIGHLIGHTED = 'NOT_HIGHLIGHTED',
+  SUCCESSFUL = 'SUCCESSFUL',
+}
+
 const ICON_PROPS: IconProps = {
   color: 'white',
   size: 24,
@@ -89,12 +95,12 @@ export function MajorMilestone({
     const { isHighlighted, status } = milestone
 
     if (!isHighlighted) {
-      return 'NOT_HIGHLIGHTED'
+      return MajorMilestoneStatus.NOT_HIGHLIGHTED
     }
     if (isEnabled && status !== TimelinePlotPointStatus.FAILED) {
-      return 'SUCCESSFUL'
+      return MajorMilestoneStatus.SUCCESSFUL
     }
-    return 'NOT_ENABLED_OR_FAILED'
+    return MajorMilestoneStatus.NOT_ENABLED_OR_FAILED
   }, [isEnabled, milestone])
 
   return (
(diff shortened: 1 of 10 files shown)


src/components/ui/timeline/majorMilestone.tsx
```diff
@@ -80,18 +85,34 @@ export function MajorMilestone({
     return { pointStyles, titleWrapperStyle }
   }, [isMobile, milestone.isHighlighted, milestone.positionPercent])
 
+  const { isNotEnabledOrHasFailed, isNotHighlighted, isSuccessful } = useMemo(() => {
+    const { isHighlighted, status } = milestone
+
```
nrenard, line 90: np: It seems like there's no chance that more then one will be true the same time. Do we need to calculate 3 of them every time? Or we can return a constant like `milestoneStatus` and then use that as reliable information?

src/components/ui/timeline/timeline.tsx
```diff
@@ -118,8 +118,7 @@ export function Timeline(props: TimelineProps) {
       {majorMilestones.map((majorMilestone, index) => (
         <MajorMilestone
           countryCode={props.countryCode}
-          isFirstMilestone={index === 0}
-          isHighlightEnabled={Math.floor(majorMilestone.positionPercent) <= animation.progress}
+          isEnabled={Math.floor(majorMilestone.positionPercent) <= animation.progress}
```
theo-nejm, line 121: can we move it to a variable and import it here?

theo-nejm: maybe add it to the majorMilestone object itself?

src/components/ui/timeline/timeline.tsx
```diff
@@ -129,7 +128,7 @@ export function Timeline(props: TimelineProps) {
       {minorMilestones.map((minorMilestone, index) => (
         <MinorMilestone
           countryCode={props.countryCode}
-          isHighlightEnabled={Math.floor(minorMilestone.positionPercent) <= animation.progress}
+          isEnabled={Math.floor(minorMilestone.positionPercent) <= animation.progress}
```
theo-nejm, line 131: same as the above comment

src/components/ui/timeline/majorMilestone.tsx
```diff
@@ -23,6 +23,12 @@ interface MajorMilestoneProps {
   milestone: Milestone
 }
 
+enum MajorMilestoneStatus {
+  NOT_ENABLED_OR_FAILED = 'NOT_ENABLED_OR_FAILED',
```
theo-nejm, line 27: np: shouldn't be better "disabled" instead of "not_enabled" and what if we make 2 different status, one for disabled and other for failed and style it inside the component, even if the styles are the same? sounds a little weird a state variable with "OR", it looks like more of an implementation detail.

theo-nejm (approved):

PR #2579, opened 2025-08-08
Fix/refactor: reduce layout shift on referrals and improve code

theo-nejm:
closes #2557 

### Improving layout shifting on referrals page

Unfortunately, it is not possible to fix entirely the layout shift without a re-design on this page, because the user session can only be validated on client-side, which means when the page is hydrated, the signed in session will pop.
**Even so**, the UX can be improved and the shift can be less harmful. The strategy used is to:

- Add a loading skeleton to the referral link (that just appeared out of nowhere before)
- Fixed the select size to remove completely the flick on the skeleton > component transition

Now, everything pops at the exact same time to the screen, which means the sensation of shaking is not felt anymore. The screen flicks once, then stops.

**Key wins**

* Zeroed the layout shift in the signed-off screen (most of the cases) - this case used to be a small shift, but it existed.

Before | After 
---|---
image | image
<sup>The other metrics can be slightly different because of the environment</sup>

* Reduced considerably the layout shifting on the worst case (signed in)

Before | After 
---|---
imageimage | imageimage

**Frame-by-frame updates**

Before | After 
---|---
image | image

The other possible changes to enhance even more the flickering needs copy changes and design changes, which means that some discussion on these topics are needed. 

### Refactoring most of the rankings components

> Reduced from 90% code duplication to 5-10% (depending on the component)

> less ~40% code (to only 3 countries, the bigger the countries count, the bigger the benefit)

* Centralized the logic on commons and passed configs instead of duplicating the whole logic.
* The main changes were just labels, so now we're mostly passing it as props.
* We still exporting the base components, so if it's needed to  change the behavior for some reason, it's possible.

Pros: 
- Much easier to implement new countries.
- Less code to write on implementation.
- Change once, reflect everywhere. (For bugs, new features, etc)

Cons:
- Slightly increases the bundle size
- The base component is used everywhere, so it's a little harder to change specific behavior on specific countries - it's solved by using the base components instead of the generic one (both are exported from common files)

## Notes for reviewers

I would suggest reviewing by commits - I've committed every single context separately, so it's easier to review by context.
Every commit is self-contained, it contains the whole refactor and a single context.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/src/components/app/pageReferrals/ca/leaderboard.tsx
+++ b/src/components/app/pageReferrals/ca/leaderboard.tsx
@@ -1,9 +1,7 @@
-'use client'
-
 import { AdvocatesLeaderboard } from '@/components/app/pageReferrals/common/leaderboard'
 import { DistrictRankingEntryWithRank } from '@/utils/server/districtRankings/upsertRankings'
 import { getCAProvinceOrTerritoryNameFromCode } from '@/utils/shared/stateMappings/caProvinceUtils'
-import { COUNTRY_CODE_TO_LOCALE, SupportedCountryCodes } from '@/utils/shared/supportedCountries'
+import { SupportedCountryCodes } from '@/utils/shared/supportedCountries'
 
 interface CaAdvocatesLeaderboardProps {
   data: DistrictRankingEntryWithRank[]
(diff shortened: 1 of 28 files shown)


src/components/app/pageReferrals/common/yourLocationRanking.tsx
```diff
@@ -36,3 +47,122 @@ export function YourLocationRanking(props: YourLocationRankingProps) {
...
+  countryCode: SupportedCountryCodes
+  placeholder: string
+  title: string
+  notFoundMessage: string
+  notFromCountryMessage: string
+  formatLabel: (stateName: string, zoneName: string) => string
+  getStateName: (stateCode: string) => string
+}
+
+export function createYourLocationRanking(config: YourLocationRankingConfig) {
```
lucasrmp-cb, line 61: Since this component will be client side, wouldn't a more standard way to design it be a context on a top level component containing the config? and child components consuming it?

My pushback is that this is quite unusual and if possible I'd like to keep the config as a component's lifecycle since it can possibly be mutable (nothing stops this function from being called from inside a component)

src/components/app/pageReferrals/ca/leaderboard.tsx
```diff
@@ -15,21 +13,13 @@ export function CaAdvocatesLeaderboard(props: CaAdvocatesLeaderboardProps) {
...
-        >
-          <AdvocatesLeaderboard.Row.Label>{`${getCAProvinceOrTerritoryNameFromCode(entry.state)} - ${entry.district}`}</AdvocatesLeaderboard.Row.Label>
-        </AdvocatesLeaderboard.Row>
-      ))}
-    </AdvocatesLeaderboard>
+    <AdvocatesLeaderboard
+      countryCode={countryCode}
+      data={data}
+      formatLabel={entry =>
+        `${getCAProvinceOrTerritoryNameFromCode(entry.state)} - ${entry.district}`
```
nrenard, line 20: np: I see same behaivor for AU, can we transform that in a helper function, and reuse it?

theo-nejm: Hey @nrenard! I though a little about that and it will be as complicated as it is rn, bc the state name transform depends on the country as well. the helper would be country specific as well. the formatLabel itself doesn't repeat throughout the code.

src/components/app/pageReferrals/ca/yourConstituencyRanking.tsx
```diff
@@ -1,122 +1,32 @@
...
-  return (
-    <DefaultPlacesSelect placeholder="Enter your address" title="Your constituency" {...props} />
-  )
+const config: YourLocationRankingConfig = {
+  countryCode: SupportedCountryCodes.CA,
+  placeholder: 'Enter your address',
+  title: 'Your constituency',
+  notFoundMessage: 'Constituency not found, please try a different address.',
+  notFromCountryMessage: `Looks like your address is not from Canada, so it can't be used to filter`,
+  formatLabel: (stateName: string, zoneName: string) => `${stateName} - ${zoneName}`,
```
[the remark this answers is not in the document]
theo-nejm: same as the above

PR #2587, opened 2025-08-13
fix: bill details design

luiz-from-delta:
This PR addresses some issues related to font weight inconsistencies and timeline display on mobile devices.

The goal is to improve the overall UX and ensure a more visually appealing interface.

Changes:
- Font Weight Adjustments:
Standardized font weights across the sections of the bill details page to enhance readability.
Ensured that headings and body text maintain a consistent hierarchy.
- Timeline Improvements:
Resolved layout issues affecting the timeline display on mobile screens.
Optimized spacing and alignment for better visual flow and user interaction.


src/components/app/pageBillDetails/partials/voteSection/dtsiAvatarBox.tsx
```diff
@@ -55,11 +55,12 @@ export function DTSIAvatarBox(props: DTSIAvatarBoxProps) {
     >
       <motion.div
         animate={{ opacity: 1, y: 0 }}
+        className="flex flex-col gap-2"
         initial={{ opacity: 0.33 }}
         key={person.slug}
         transition={{ duration: 0.75 }}
       >
-        <div className="relative mx-auto w-fit">
+        <div className="relative mx-auto w-fit" style={{ height: AVATAR_SIZE }}>
```
nrenard, line 63: np: Doesn't it work like that?
```suggestion
        <div className="relative mx-auto w-fit h-fit">
```

src/components/ui/timeline/timeline.tsx
```diff
@@ -42,46 +43,78 @@ export function Timeline(props: TimelineProps) {
         return a.date.getTime() - b.date.getTime()
       })
 
-    const majorMilestones = plotPoints
-      .filter(plotPoint => plotPoint.isMajorMilestone)
+    const initialMajorMilestones = plotPoints
```
nrenard, line 46: np: What do you think about move this `useMemo` to a custom hook it will be better to maintenance and further improvements?
Also will be good add some comments for easier understand.

PR #2623, opened 2025-08-26
feat: add webpack-retry-chunk-load-plugin for improved chunk loading …

nrenard:
## Description
This PR addresses persistent Hydration issues in client side related to bad connection or unavailable server.
[PROD-SWC-WEB-12R](https://stand-with-crypto.sentry.io/issues/5018485674/events/481104b8941d44358ba61b640dc7a683/?project=4506490717470720)

fixes PROD-SWC-WEB-12R

## What changed? Why?
- Integrated RetryChunkLoadPlugin into webpack configuration to handle chunk loading failures with a maximum of 3 retries and a delay of 1000ms.

## How has it been tested?

- [x] Locally
- [x] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test

### DIFF
--- a/package.json
+++ b/package.json
@@ -206,6 +206,7 @@
     "tsconfig-paths": "^4.2.0",
     "tsx": "^4.7.0",
     "typescript": "^5.7.2",
+    "webpack-retry-chunk-load-plugin": "^3.1.1",
     "yargs": "^17.7.2"
   },
   "overrides": {
--- a/next.config.ts
+++ b/next.config.ts
@@ -1,6 +1,7 @@
 import bundleAnalyzer from '@next/bundle-analyzer'
 import { withSentryConfig } from '@sentry/nextjs'
 import type { NextConfig } from 'next'
+import { RetryChunkLoadPlugin } from 'webpack-retry-chunk-load-plugin'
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -206,6 +206,7 @@
     "tsconfig-paths": "^4.2.0",
     "tsx": "^4.7.0",
     "typescript": "^5.7.2",
+    "webpack-retry-chunk-load-plugin": "^3.1.1",
```
theo-nejm, line 209: Hey @nrenard, it looks like the content of this lib you're adding on this PR is a single file with around 100 lines:
https://github.com/mattlewis92/webpack-retry-chunk-load-plugin/blob/master/src/index.ts

The last update was 3 years ago. Does that make sense to add this file as a util file on our current structure instead of installing a new lib? Looks like a simple copy paste task. Circle back if that makes sense.

nrenard: Good point about the maintenance status! 
I agree we can copy the code if/when we need to make changes or if the package becomes problematic. For now, since it's working and we're just testing the approach, I think the npm dependency is fine. We can always vendor it later if we run into issues. What do you think?

theo-nejm (approved):

PR #2689, opened 2025-09-17
feat: add top level client to eu + refactor navbar and footer config structure

theo-nejm:
## What changed? Why?

This PR adds the top level client (with the pages - for now without content). And the layout component for the EU pages.

For this to work properly, this PR has also a refactor on navbar and footer configuration structure, changing from a static object to a function to dynamically set some props and i18n texts based on the language and maintain the compatibility with the builder.io pages as well.

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/utils/server/metadataUtils.ts
```diff
@@ -109,3 +109,11 @@ export function generateCountryCodeLayoutMetadata(countryCode: SupportedCountryC
     ...generateTopLevelMetadataDetails(countryCode),
   } satisfies Metadata
 }
+
+export const generateCountryCodeLayoutMetadataWithLanguage = (
+  countryCode: SupportedCountryCodes,
+  language: string,
```
twistershark, line 115: might be worth making this more strict

PR #2762, opened 2025-10-07
feature: create profile page on EU

theo-nejm:
closes #2756

## What changed? Why?

- Added i18n messages for user profile forms, email and SMS subscription forms, and various UI components to support multiple languages (EN, DE, FR).
- Updated components to utilize translation functions for dynamic text rendering, improving localization and user experience.
- Refactored existing components to remove hardcoded strings and replace them with translatable keys, ensuring consistency across the application.
- Enhanced accessibility by providing translated titles and descriptions for user actions and forms.

## How has it been tested?

- [x] Locally
- [ ] Vercel Preview Branch
- [ ] Unit test
- [ ] Functional test


src/components/app/pageUserProfile/common/index.tsx
```diff
@@ -110,11 +193,11 @@ export function PageUserProfile({
                 {getSensitiveDataUserDisplayName(user)}
               </div>
               <div className="text-sm text-gray-500">
-                Joined{' '}
+                {t('joined')}{' '}
                 <FormattedDatetime
                   date={new Date(user.datetimeCreated)}
                   dateStyle="medium"
-                  locale={COUNTRY_CODE_TO_LOCALE[countryCode]}
+                  locale={getLocaleForLanguage(language)}
```
lucasrmp-cb, line 200: Don't know in case of french and german so we don't need to change. But some languages will swap the order of these elements so a good practice is never to translate parts of a phrase and instead use interpolations so the whole phrase can be manipulated by the language as necessary

src/components/app/updateUserProfileForm/step1/index.tsx
```diff
@@ -153,11 +230,10 @@ export function UpdateUserProfileForm({
       >
         <div>
           <PageTitle className="mb-1" size="md">
-            {hasCompleteUserProfile(user) ? 'Edit' : 'Finish'} your profile
+            {hasCompleteUserProfile(user) ? t('edit') : t('finish')} {t('yourProfile')}
```
lucasrmp-cb, line 233: same here for `t('finish')} {t('yourProfile')`, some languages will swap the order here (I know japanese would, for sure there are other examples), and even though we're not supporting those languages It's not a good practice

src/components/app/updateUserProfileForm/step1/privacyConsentDisclaimer.tsx
```diff
@@ -2,21 +2,46 @@ import { Collapsible, CollapsibleContent } from '@/components/ui/collapsible'
...
     <Collapsible open={shouldShowConsentDisclaimer}>
       <CollapsibleContent className="AnimateCollapsibleContent">
         <FormDescription className="text-center md:text-left">
-          By completing your profile, you consent to and understand that SWC and its vendors may
-          collect and use your personal information subject to{' '}
-          <InternalLink href={intlUrls.privacyPolicy()}>SWC Privacy Policy</InternalLink>.
+          {t('privacyConsentDisclaimerPart1')}{' '}
+          <InternalLink href={intlUrls.privacyPolicy()}>
+            {t('privacyConsentDisclaimerPart2')}
+          </InternalLink>
```
lucasrmp-cb, line 43: Here is more of an api question for me. Do we support something like [this](https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0)?

lucasrmp-cb (approved):

src/components/app/updateUserProfileForm/step1/privacyConsentDisclaimer.tsx
theo-nejm: We can create this component, but we don't support that atm.

